PR #4, opened 2016-12-13
Updating the timestamp to a more condenced format.

mbruzek:
Resolves https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/147

We need to land these changes before our automated system can build with this patch, unless I change the job to accept branches, which I can do that is fine.


chuckbutler:
neat fix :) I like it

+1 LGTM

PR #9, opened 2017-03-27
add a script for building k8s snaps

wwwtyro:


jenkins-build-k8s-snaps.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+KUBE_VERSION=${KUBE_VERSION:-"v1.5.2"}
```
Cynerva: `v1.5.2` seems like an odd default to have. Why this?

jenkins-build-k8s-snaps.sh
```diff
@@ -0,0 +1,6 @@
...
+
+git clone https://github.com/juju-solutions/release.git --branch rye/snaps
```
Cynerva: Hey as long as we're checking out a branch directly, let's make it a shallow clone with `--depth 1`

jenkins-build-k8s-snaps.sh
wwwtyro: Only because I saw it elsewhere: https://github.com/juju-solutions/kubernetes-jenkins/blob/f1587326589b48d6d1370972da0f57335638e21d/build-kubernetes.sh#L15

Cynerva:
Gonna try squash and merge on this one. Why not?

PR #12, opened 2017-04-27
push snaps to edge

wwwtyro:


build-and-release-k8s-snaps.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -eux
+
+git clone https://github.com/juju-solutions/release.git --branch rye/snaps --depth 1
+cd release/snap && make KUBE_VERSION=$KUBE_VERSION
+
+for app in kubeadm kube-apiserver kube-controller-manager kubectl kubefed kubelet kube-proxy kube-scheduler; do
+  snapcraft push build/${app}_${KUBE_VERSION:1}_amd64.snap --release ${KUBE_VERSION:1:3}/edge
```
Cynerva: We'll need to figure out when we should or should not release to latest/edge as well.

Cynerva: We'll need to add kubernetes-test here soon. Don't have the 1.6/1.7 tracks made yet though so it can wait.

PR #17, opened 2017-04-28
disable ppc

wwwtyro:


Cynerva:
For context, ppc64le is causing us to fail during the `snapcraft push`. The snap build will need to be updated to map ppc64le -> ppc64el as the snap architecture.

PR #29, opened 2017-05-09
Fix gubernator timestamp formatting

chuckbutler:
The old format had prefixing 0's or other special chars in history that were preventing dashboard from ingesting.


vorlonofportland:
This is not the typical format used by upstream, but it matches what we're doing for Ubuntu in GKE and looks good to me.

PR #30, opened 2017-05-11
Delete unused scripts

Cynerva:
Cleaning up some old junk that we're not using anymore.

Some of these are still referenced in existing jobs (e.g. kubernetes-charm-test-e2e calls jenkins-charm-e2e.sh) but the intention is to remove those jobs as well.


ktsakalozos:
LGTM +1

PR #31, opened 2017-05-11
move snap scripts to snaps/ dir

Cynerva:
Organizing snap scripts into a snaps/ subdir. I'll update the jobs to point to the right place when this is merged.


ktsakalozos:
LGTM +1

PR #35, opened 2017-05-12
Make the build-and-release-all script work

Cynerva:
Making the build-and-release-all.sh script work. This brings in functionality that was previously covered by the big-red-button job.

### DIFF
--- a/charms/build-and-release-bundles.sh
+++ b/charms/build-and-release-bundles.sh
@@ -19,3 +19,25 @@ PUSH_CMD="/usr/bin/charm push ./bundles/core-flannel ${CORE}"
 CORE_REVISION=`${PUSH_CMD} | tail -n +1 | head -1 | awk '{print $2}'`
 /usr/bin/charm release --channel edge ${CORE_REVISION}
 /usr/bin/charm grant --channel edge ${CORE_REVISION} everyone
+
+if [ "$RUN_TESTS" = "true" ]; then
+  export CHANNEL="edge"
+  export RUN_MATRIX_TESTS=true
+  export BUNDLE_NAME="canonical-kubernetes"
+  export CLOUD=google
+  ./tests/test-bundle.sh
+  export CLOUD=aws
+  ./tests/test-bundle.sh
+  export BUNDLE_NAME="kubernetes-core"
+  export CLOUD=google
+  ./tests/test-bundle.sh
+  export CLOUD=aws
+  ./tests/test-bundle.sh
+
+  export FROM_CHANNEL=edge
+  export TO_CHANNEL=beta
+  export BUNDLE_NAME_AND_REVISION=$(charm show cs:~containers/bundle/canonical-kubernetes --channel edge id | grep Id | awk '{print $2}')
+  ./charms/release-bundle-to-channel.sh
+  export BUNDLE_NAME_AND_REVISION=$(charm show cs:~containers/bundle/kubernetes-core --channel edge id | grep Id | awk '{print $2}')
+  ./charms/release-bundle-to-channel.sh
+fi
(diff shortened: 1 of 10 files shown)


charms/build-and-release-easyrsa.sh
```diff
@@ -48,6 +44,6 @@ touch report.xml
 if [ ${RELEASE} = true ]; then
   CHARM=$(/usr/bin/charm push $JUJU_REPOSITORY/builds/easyrsa cs:~containers/easyrsa | head -n 1 | awk '{print $2}')
   echo "Releasing ${CHARM}"
-  charm release ${CHARM} --channel ${RELEASE_TO_CHANNEL} -r easyrsa-${RESOURCE_REV}
+  charm release ${CHARM} --channel ${RELEASE_TO_CHANNEL} -r easyrsa-5
```
mbruzek: Do you really want "-5" here? Will that not cause a problem?

charms/build-and-release-flannel.sh
```diff
@@ -48,6 +44,6 @@ touch report.xml
 if [ ${RELEASE} = true ]; then
   CHARM=$(/usr/bin/charm push $JUJU_REPOSITORY/builds/flannel cs:~containers/flannel | head -n 1 | awk '{print $2}')
   echo "Releasing ${CHARM}"
-  charm release ${CHARM} --channel ${RELEASE_TO_CHANNEL} -r flannel-${RESOURCE_REV}
+  charm release ${CHARM} --channel ${RELEASE_TO_CHANNEL} -r flannel-9
```
mbruzek: Hard coding the revision number seems like a bad idea.

charms/build-and-release-easyrsa.sh
Cynerva: Yeah, this is a short-term step back but I think it's a move in the right direction.

We currently surface the RESOURCE_REV as a job parameter in Jenkins, meaning the user has to specify it when running a job. That's error-prone, especially when we're building all the charms as part of a single job. We've had problems with messed up revision numbers in a few of our releases already.

Hard-coding has obvious problems of its own, as you're right to point out. Updating resources will require a commit to bump the revs. We don't want that kind of friction in our release process.

I made this change to localize the badness, if that makes sense. Passing an error-prone parameter through three scripts and a GUI just doesn't feel right.

What we really need is to have our charm build scripts always grab the latest revisions, or alternatively, build and push their own resources along with the charm.

Cynerva:
Removed hardcoded resources across the board. Added functionality to grab the latest resource revisions instead.

chuckbutler:
I like the current iteration much better than the hardcoded paths. 👍

charms/build-and-release-bundles.sh
```diff
@@ -19,3 +19,25 @@ PUSH_CMD="/usr/bin/charm push ./bundles/core-flannel ${CORE}"
...
+  ./tests/test-bundle.sh
+  export BUNDLE_NAME="kubernetes-core"
+  export CLOUD=google
+  ./tests/test-bundle.sh
+  export CLOUD=aws
+  ./tests/test-bundle.sh
+
+  export FROM_CHANNEL=edge
+  export TO_CHANNEL=beta
+  export BUNDLE_NAME_AND_REVISION=$(charm show cs:~containers/bundle/canonical-kubernetes --channel edge id | grep Id | awk '{print $2}')
```
chuckbutler: What happens to this when the charm store stops responding or spits out an error message? Should we have some kind of validator here to halt running or is this a set -e good enough to catch that?

chuckbutler: same on line 42

charms/build-and-release-easyrsa.sh
```diff
@@ -1,11 +1,7 @@
 #!/usr/bin/env bash
-# Runs the charm build for the Canonical Kubernetes charms.
-
-set -o errexit  # Exit when an individual command fails.
-set -o nounset  # Exit when undeclaried variables are used.
-set -o pipefail  # The exit status of the last command is returned.
-set -o xtrace  # Print the commands that are executed.
+set -eux
```
chuckbutler: <3

PR #36, opened 2017-05-15
Fix wrong working directory after charm builds

Cynerva:
Some builds are failing because we don't restore the working directory properly after the charm build. This should fix that.


chuckbutler:
Ah I didn't catch that. Nice save! ⚽️

PR #37, opened 2017-05-16
Promote all charms and bundles together, rather than by bundle

Cynerva:
This is to fix the following problems:
1. We still had hard-coded resources in `release-bundle-to-channel.sh`
2. We were never promoting kubernetes-e2e

Changing our strategy a bit here by promoting all of the charms and bundles together, rather than doing it by bundle.

@chuckbutler @ktsakalozos @wwwtyro


charms/build-and-release-easyrsa.sh
```diff
@@ -44,6 +44,5 @@ touch report.xml
 if [ ${RELEASE} = true ]; then
   CHARM=$(/usr/bin/charm push $JUJU_REPOSITORY/builds/easyrsa cs:~containers/easyrsa | head -n 1 | awk '{print $2}')
   echo "Releasing ${CHARM}"
-  ./charms/charm-release-with-latest-resources.sh ${CHARM} --channel ${RELEASE_TO_CHANNEL}
-  charm grant ${CHARM} everyone --channel ${RELEASE_TO_CHANNEL}
+  CHARM="$CHARM" FROM_CHANNEL=unpublished TO_CHANNEL=edge ./charms/promote-charm.sh
```
chuckbutler: did we want this to always only go to edge?

Cynerva: Oh yeah, forgot about this. This removes a `RELEASE_TO_CHANNEL` param that we pass in from Jenkins.

I'm asserting that the automation for individual charms should only ever push to edge. We can promote from there separately.

charms/promote-all-charms-and-bundles.sh
```diff
@@ -0,0 +1,18 @@
...
+  cs:~containers/kubernetes-master
+  cs:~containers/kubernetes-worker
+  cs:~containers/kubeapi-load-balancer
+  cs:~containers/kubernetes-e2e
+  cs:~containers/kubernetes-core
+  cs:~containers/canonical-kubernetes
+"
+
+for charm in $CHARMS; do
+  CHARM="$charm" FROM_CHANNEL="$FROM_CHANNEL" TO_CHANNEL="$TO_CHANNEL" ./charms/promote-charm.sh
```
chuckbutler: And do we want this to be from channel to channel or do we want it to be revision # to channel?

Cynerva: We'd definitely need `FROM_CHANNEL` either way. Resource revisions can differ depending on which channel you get it from, even if the charm revision is the same.

So when we promote from candidate to stable, for example, we better make sure we're getting the resources from candidate.

I'm also trying to make sure we don't have to put in manual revision #s since we've had problems with that in the past. The automation can use them as needed (and promote-charm.sh does use them) but we gotta do our best to keep that internal, IMO.

chuckbutler:
LGTM, we can adjust as required.

PR #38, opened 2017-05-17
Don't do `charm grant` after every release

Cynerva:


ktsakalozos:
Probably my misunderstanding. I thought this would not change anything. 

LGTM +1

PR #39, opened 2017-05-25
Refactor bundle test scripts to have a single entry point

Cynerva:
Bit of a refactor here, with the following goals:

1. Standardize on a single entry point for running bundle tests (replaces test-bundle.sh and test-bundle-snap-combination.sh)
2. Allow build-and-release-bundles.sh to test against edge snaps
3. Remove gce from our default set of tests, for now, since the failures we're getting there aren't meaningful to CDK


chuckbutler:
this LGTM.

Curious though, what test failures were happening in GCE?

Cynerva:
Thanks. Re: GCE, it's a failure during test setup when amulet tries to upload scripts to the units. sshguard shenanigans. Bugs: https://github.com/juju/amulet/issues/170 and https://bugs.launchpad.net/juju/+bug/1669501

PR #44, opened 2017-06-06
Initial upgrade tests

Cynerva:
This is a new standalone test suite, using pytest and libjuju, for doing upgrade testing against CDK.

The tests follow this pattern:
1. Deploy initial bundle
2. Upgrade the deployment in some way (charms or snaps)
3. Assert that the deployment is healthy

I'm still thinking about step 3, the health check, and what it should look like. It's pretty weak for now but I intend to improve it.

### DIFF
--- a/tests/upgrade-tests/test_upgrade_snaps.py
+++ b/tests/upgrade-tests/test_upgrade_snaps.py
@@ -4,10 +4,10 @@ from validation import validate_all
 
 test_cases = [
     # bundle                 charm_channel  from_channel  to_channel
-    ('kubernetes-core',      'beta',        '1.6/stable', '1.6/edge'),
-    ('kubernetes-core',      'beta',        '1.6/stable', '1.7/edge'),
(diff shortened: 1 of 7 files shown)


tests/upgrade-tests/test_upgrade_snaps.py
```diff
@@ -0,0 +1,33 @@
...
+    await master.set_config({'channel': channel})
+    worker = model.applications['kubernetes-worker']
+    await worker.set_config({'channel': channel})
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize('bundle,from_channel,to_channel', test_cases)
+async def test_upgrade_snaps(bundle, from_channel, to_channel):
+    async with temporary_model() as model:
+        await model.deploy(bundle, channel='stable')
```
Cynerva: This should probably be `channel='edge'`. I'm still thinking about it but if anyone has any opinions, lemme know!

PR #46, opened 2017-06-09
Validate snap versions are consistent with snap channel config

Cynerva:
Mainly adding this to get some confidence that test_upgrade_snaps is actually upgrading the snaps properly.

### DIFF
--- a/tests/upgrade-tests/utils.py
+++ b/tests/upgrade-tests/utils.py
@@ -40,7 +40,11 @@ async def temporary_model():
 
 def assert_no_unit_errors(model):
     for unit in model.units.values():
-        assert unit.data['workload-status']['current'] != 'error'
+        workload_status = unit.data['workload-status']['current']
+        juju_status = unit.data['juju-status']['current']
+        assert workload_status != 'error'
+        assert juju_status != 'failed'
+        assert juju_status != 'lost'
 
 
 def all_units_ready(model):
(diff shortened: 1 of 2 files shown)


tests/upgrade-tests/utils.py
```diff
@@ -40,7 +40,11 @@ def dump_model_info(model):
 
 def assert_no_unit_errors(model):
     for unit in model.units.values():
-        assert unit.data['workload-status']['current'] != 'error'
+        workload_status = unit.data['workload-status']['current']
```
chuckbutler: nice :+1:

tests/upgrade-tests/validation.py
```diff
@@ -18,6 +19,40 @@ def validate_status_messages(model):
...
+            'kube-scheduler',
+            'cdk-addons',
+        ],
+        'kubernetes-worker': [
+            'kubectl',
+            'kubelet',
+            'kube-proxy',
+        ],
+    }
+    for app_name, snaps in snaps_to_validate.items():
```
chuckbutler: I would caution that we add some comment of the output snap list is expected to return so that when it breaks and our screen scrape parser misbehaves we know why. I ran into this with the layer-etcd snap refactoring.

Cynerva: Good idea, I'll add that

chuckbutler:
Aces :beers:

PR #47, opened 2017-06-09
Fix error check failing because 'juju-status' does not exist

Cynerva:
Sometimes the data comes back with no `'juju-status'` key.

I haven't tested this fix yet, but will kick off a test and report back on Monday.

### DIFF
--- a/tests/upgrade-tests/utils.py
+++ b/tests/upgrade-tests/utils.py
@@ -41,10 +41,11 @@ async def temporary_model():
 def assert_no_unit_errors(model):
     for unit in model.units.values():
         workload_status = unit.data['workload-status']['current']
-        juju_status = unit.data['juju-status']['current']
         assert workload_status != 'error'
-        assert juju_status != 'failed'
-        assert juju_status != 'lost'
+        if 'juju-status' in unit.data:
+            juju_status = unit.data['juju-status']['current']
+            assert juju_status != 'failed'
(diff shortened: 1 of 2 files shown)


tests/upgrade-tests/utils.py
```diff
@@ -41,10 +41,11 @@ def dump_model_info(model):
 def assert_no_unit_errors(model):
     for unit in model.units.values():
         workload_status = unit.data['workload-status']['current']
-        juju_status = unit.data['juju-status']['current']
         assert workload_status != 'error'
-        assert juju_status != 'failed'
-        assert juju_status != 'lost'
+        if 'juju-status' in unit.data:
```
ktsakalozos: This seems to silence the error. Could this be shadowing any real problem? Do you fancy a retry before giving up with a message in the logs saying we gave up because we did not get juju status?

tvansteenburgh: In juju 2+, 'juju-status' has become 'agent-status'. You can use `unit.agent_status` here. And `unit.workload_status` above. See https://github.com/juju/python-libjuju/blob/master/juju/unit.py#L13 and https://github.com/juju/python-libjuju/blob/master/juju/unit.py#L34.

Cynerva:
Oops, looks like I double-botched this one. Yay!

Put up another set of fixes to use the unit properties throughout. Thanks for the tip @tvansteenburgh

PR #54, opened 2017-06-22
Wrap tests in a large timeout

Cynerva:
This introduces a timeout at the `temporary_model` level. Hopefully this will prevent tests from stalling out forever.

Also removed the timeout from `wait_for_ready` in favor of the higher level timeout. Seems nicer to have just a single, nearly-global timeout per test.

### DIFF
--- a/tests/upgrade-tests/utils.py
+++ b/tests/upgrade-tests/utils.py
@@ -5,6 +5,7 @@ import sys
 import logging
 from asyncio_extras import async_contextmanager
 from async_generator import yield_
+from contextlib import contextmanager
 from juju.controller import Controller
 from juju.model import Model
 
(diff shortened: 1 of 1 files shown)


tests/upgrade-tests/utils.py
```diff
@@ -45,26 +46,45 @@ def dump_model_info(model):
...
+    '''
+    loop = asyncio.get_event_loop()
+    task = asyncio.Task.current_task()
+    handle = loop.call_later(timeout, task.cancel)
+    try:
+        yield
+    except asyncio.CancelledError:
+        raise asyncio.TimeoutError('Timed out after %f seconds' % timeout)
+    finally:
+        handle.cancel()
```
Cynerva: This approach seems weird but I couldn't come up with anything better. Open to suggestions!

PR #55, opened 2017-06-26
Bump upgrade test time limit to 1 hour

Cynerva:
Had another test timeout. The blocker was kubernetes-master still "Waiting for kube-system pods", but it's not clear with the current output if it was stuck there or if it just didn't have enough time.

Let's just bump it up to 1 hour for now. I'd say if we still see failures after that, then we're probably well into "something went wrong" territory.


ktsakalozos:
Master is the last charm that comes ready, but I've never seen it not coming up for no reason. Merging this since will help in our investigation.

PR #56, opened 2017-06-27
deploy via conjure-up

wwwtyro:
@Cynerva 

This still needs some TLC (refactoring and further integration), but here's a rough draft with the `test_deploy` tests working.

### DIFF
--- a/tests/upgrade-tests/utils.py
+++ b/tests/upgrade-tests/utils.py
@@ -122,7 +122,7 @@ async def wait_for_ready(model):
         assert_no_unit_errors(model)
         await asyncio.sleep(1)
 
-async def conjureup(namespace, bundle, channel, snap_channel, model):
+async def conjureup(model, namespace, bundle, channel, snap_channel=None):
     with tempfile.TemporaryDirectory() as tmpdirname:
         cmd = 'charm pull --channel=%s cs:~%s/%s %s'
         cmd %= channel, namespace, bundle, os.path.join(tmpdirname, bundle)
@@ -143,8 +143,9 @@ async def conjureup(namespace, bundle, channel, snap_channel, model):
         appkey = 'services' if 'services' in bundledata else 'applications'
         master = bundledata[appkey]['kubernetes-master']
         worker = bundledata[appkey]['kubernetes-worker']
-        master.setdefault('options', {})['channel'] = snap_channel
-        worker.setdefault('options', {})['channel'] = snap_channel
+        if snap_channel is not None:
+            master.setdefault('options', {})['channel'] = snap_channel
+            worker.setdefault('options', {})['channel'] = snap_channel
         with open(os.path.join(tmpdirname, 'spell', 'bundle.yaml'), 'w') as f:
             yaml.dump(bundledata, f, default_flow_style=False)
         with open(os.path.join(tmpdirname, 'spell', 'metadata.yaml')) as f:
--- a/tests/upgrade-tests/utils.py
+++ b/tests/upgrade-tests/utils.py
@@ -175,12 +175,12 @@ def asyncify(f):
     return wrapper
 
 
-async def deploy_e2e(model, charm_channel='stable', snap_channel='stable'):
-    await model.deploy('cs:~containers/kubernetes-e2e', channel=charm_channel)
+async def deploy_e2e(model, charm_channel='stable', snap_channel=None):
+    config = None if snap_channel is None else {'channel': snap_channel}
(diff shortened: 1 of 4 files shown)


tests/upgrade-tests/utils.py
```diff
@@ -116,3 +121,48 @@ def all_units_ready(model):
...
+        await process.wait()
+        assert process.returncode == 0
+        shutil.copytree(
+            os.path.join('/snap/conjure-up/current/spells', bundle),
+            os.path.join(tmpdirname, 'spell')
+        )
+        os.remove(os.path.join(tmpdirname, 'spell', 'steps', 'step-01_get-kubectl'))
+        os.remove(os.path.join(tmpdirname, 'spell', 'steps', 'step-01_get-kubectl.yaml'))
+        os.remove(os.path.join(tmpdirname, 'spell', 'steps', 'step-02_cluster-info'))
+        os.remove(os.path.join(tmpdirname, 'spell', 'steps', 'step-02_cluster-info.yaml'))
```
Cynerva: We probably shouldn't remove these steps. They're part of our production deployment process so they should be part of our tests, too.

tests/upgrade-tests/utils.py
```diff
@@ -116,3 +121,48 @@ def all_units_ready(model):
...
+        with open(os.path.join(tmpdirname, bundle, 'bundle.yaml')) as f:
+            bundledata = yaml.load(f)
+        appkey = 'services' if 'services' in bundledata else 'applications'
+        master = bundledata[appkey]['kubernetes-master']
+        worker = bundledata[appkey]['kubernetes-worker']
+        master.setdefault('options', {})['channel'] = snap_channel
+        worker.setdefault('options', {})['channel'] = snap_channel
```
Cynerva: We'll need to make `snap_channel` optional. If we're just testing the latest charms, for example, we probably want to stick with the snap channels that are specified in the bundle.

tests/upgrade-tests/utils.py
wwwtyro: Agreed, unfortunately they're in the way of testing because they require passwordless sudo. @stokachu, any ideas here? We want to be able to run these on developer boxes, so requiring passwordless sudo doesn't feel right.

tests/upgrade-tests/utils.py
```diff
@@ -116,3 +121,48 @@ def all_units_ready(model):
...
+        with open(os.path.join(tmpdirname, 'spell', 'bundle.yaml'), 'w') as f:
+            yaml.dump(bundledata, f, default_flow_style=False)
+        with open(os.path.join(tmpdirname, 'spell', 'metadata.yaml')) as f:
+            metadata = yaml.load(f)
+        del metadata['bundle-name']
+        with open(os.path.join(tmpdirname, 'spell', 'metadata.yaml'), 'w') as f:
+            yaml.dump(metadata, f, default_flow_style=False)
+        cmd = 'juju show-controller --format=json'.split()
+        controller_raw = subprocess.check_output(cmd)
```
Cynerva: need to make this subprocess call asyncio-friendly

tests/upgrade-tests/utils.py
battlemidget: With headless there isn't a way around this in the current code. This is for unattended deploy tests. We may can work something out like a cli argument for prompting for sudo password. @johnsca  what do you think?

johnsca: If this is for automated testing, I'm not sure that prompting for the password would help.  If the test can supply the password if prompted, could it invoke sudo to cache it like conjure-up interactive does?

Alternatively, if the tests are containerized, could they be granted passwordless sudo in a safe way?

wwwtyro: Would ignoring these steps if passwordless sudo is unavailable be an option? Perhaps print a warning - "kubectl and kubefed not installed on host"?

Bigger picture - feels a little weird that we're forcing this installation. Can it be prompted? If we can, then we can maybe skip the step if a tty isn't detected (in the headless case)?

battlemidget: We don't have a way to say "skip this spell if condition is met",  It is worth investigating though. I think we want to stay away from any prompting in the headless case as this is primarily meant to be opinionated beyond what can be customized via environment variables

wwwtyro: Gonna defer this until we have a good solution.

Cynerva:
Looking good! In it goes.

PR #58, opened 2017-06-30
Fix bug on labels

ktsakalozos:
The CI is setting the initial bundle options in the patched files. The tests were expecting the unit test's setup method to be used instead but it is not.


ktsakalozos:
Merging this PR to move to the next CI blocker.

PR #60, opened 2017-06-30
E2e validator

ktsakalozos:


tests/upgrade-tests/test_upgrade_charms.py
```diff
@@ -14,9 +14,10 @@
 @pytest.mark.parametrize('bundle,from_channel,to_channel', test_cases)
 async def test_upgrade_charms(bundle, from_channel, to_channel):
     async with temporary_model() as model:
-        await model.deploy(bundle, channel=from_channel)
+        await deploy_bundle(model, bundle, from_channel)
         await wait_for_ready(model)
         for app in model.applications.values():
             await app.upgrade_charm(channel=to_channel)
+        await add_new_e2e_relation(model)
```
Cynerva: So this test will deploy the old e2e, upgrade it, then add the new relation.

Does it make sense to upgrade e2e at all? It seems like it'd be cleaner to just deploy the newer e2e after the upgrades have taken place -- that way, all of its relations are handled in one place. On the other hand, being able to deploy e2e in advance might time in the test run.

I'm not sure where I stand on that, so we'll call it good and I'll revisit later :)

tests/upgrade-tests/utils.py
```diff
@@ -47,6 +48,14 @@ def dump_model_info(model):
     return model
 
 
+async def get_model(name):
+    ''' Get current model
+    '''
+    model = Model()
+    await model.connect_model(name)
+    return model
```
Cynerva: This looks like it's not called anywhere, can you remove it?

tests/upgrade-tests/validation.py
```diff
@@ -86,3 +88,19 @@ def validate_status_messages(model):
...
+    workers = model.applications['kubernetes-worker']
+    await workers.set_config({'allow-privileged': 'true'})
+    if len(workers.units) < 2:
+        await workers.add_unit(1)
+    await wait_for_ready(model)
+
+    e2e_unit = model.applications['kubernetes-e2e'].units[0]
+    action = await e2e_unit.run_action('test')
+    await action.wait()
+    assert action.status == 'completed'
```
Cynerva: I've gotten a ton of testing in on this validator over the past week while testing 1.7. Solid :+1:

PR #61, opened 2017-07-03
Validate kubelet auth against /pods/ instead of /runningpods/

Cynerva:
I'm getting a `301 Moved Permanently` from /runningpods/ against 1.7. Let's use /pods/ instead.

This is untested ATM.


ktsakalozos:
LGTM +1

PR #67, opened 2017-07-18
Add test_live_model.py

Cynerva:
To aid in developing validators quickly.

This wouldn't be run in CI at all, but would be usable by devs when developing validators. Gives us a way to rerun them without having to wait for a whole new fresh deployment.

I'm on the fence about this one, and it needs testing, but figured I'd throw a PR up for opinions. Is this valuable to us?


wwwtyro:
I want this, like, yesterday. Can I merge? Won't break anything as is, soo...

Cynerva:
Sure, go for it

PR #68, opened 2017-07-18
add microbot validation

wwwtyro:
tested :+1:

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -74,8 +74,12 @@ async def validate_microbot(model):
     action = await unit.run_action('microbot', replicas=3)
     await action.wait()
     assert action.status == 'completed'
-    # TODO: wait for pods running
-    # TODO: test that we can reach the ingress endpoint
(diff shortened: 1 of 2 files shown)


Cynerva:
Looks good, thanks!

integration-tests/validation.py
```diff
@@ -74,8 +74,12 @@ def validate_status_messages(model):
     action = await unit.run_action('microbot', replicas=3)
     await action.wait()
     assert action.status == 'completed'
-    # TODO: wait for pods running
-    # TODO: test that we can reach the ingress endpoint
+    for i in range(60):
+        resp = requests.get('http://' + action.data['results']['address'])
```
tvansteenburgh: This blocks the event loop. It would be better to use aiohttp, or run this with `run_in_executor`.

wwwtyro: :+1: I'll be in here again shortly, I'll sneak this in.

PR #70, opened 2017-07-19
Update charm builds to run the new tests

Cynerva:
This updates build-and-release-bundles (and build-and-release-all-charms-and-bundles) to run the new integration tests.

This also removes the old bundle test runner, since a test case for that is included in the integration tests.

Checking this in will break the Jenkins jobs until they've been updated to account for differences here.

It's also worth noting that this will make the build-and-release-all-charms-and-bundles jobs quite a bit slower -- maybe 3-4 hours total run time, instead of the 1.5 hours of today.


ktsakalozos:
LGTM +1

PR #71, opened 2017-07-19
Add canonical-kubernetes-elastic to promote-all-charms-and-bundles

Cynerva:
Wow, I forgot to do this. It was even called out in our trello card:

> We also need to update promote-all-charms-and-bundles job


ktsakalozos:
LGTM +1

PR #72, opened 2017-07-20
add dashboard test

wwwtyro:
This checks that:
 - the dashboard url returns 200
 - a call to the dashboard api returns 200
 - ~there are metrics returned in the api call~

~The metrics are checked by making sure there's >0 cpu and memory data points. This is a little risky because there simply may not be any yet. To address this, I've placed the dashboard check after the e2e testing.~

NVM, metrics checking is too flaky. Getting rid of it for now.

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -82,7 +82,7 @@ async def validate_microbot(model):
     assert action.status == 'completed'
     for i in range(60):
         resp = await asyncify(requests.get)('http://' + action.data['results']['address'])
-        if resp.ok:
+        if resp.status_code == 200:
             return
         await asyncio.sleep(1)
     raise MicrobotError('Microbot failed to start.')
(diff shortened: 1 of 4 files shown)


integration-tests/validation.py
```diff
@@ -75,13 +81,37 @@ def validate_status_messages(model):
     await action.wait()
     assert action.status == 'completed'
     for i in range(60):
-        resp = requests.get('http://' + action.data['results']['address'])
+        resp = await asyncify(requests.get)('http://' + action.data['results']['address'])
```
wwwtyro: @tvansteenburgh snuck 'er in

integration-tests/validation.py
```diff
@@ -75,13 +81,37 @@ def validate_status_messages(model):
...
+    with NamedTemporaryFile() as f:
+        await unit.scp_from('config', f.name)
+        with open(f.name, 'r') as stream:
+            config = yaml.load(stream)
+    url = config['clusters'][0]['cluster']['server']
+    user = config['users'][0]['user']['username']
+    password = config['users'][0]['user']['password']
+    auth = requests.auth.HTTPBasicAuth(user, password)
+    resp = await asyncify(requests.get)(url, auth=auth, verify=False)
+    assert resp.ok is True
```
Cynerva: Can you change this to `assert resp.status_code == 200`?

http://docs.python-requests.org/en/master/api/#requests.Response.ok

integration-tests/validation.py
```diff
@@ -75,13 +81,37 @@ def validate_status_messages(model):
...
+    url = '%s/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/api/v1/workload/default?filterBy=&itemsPerPage=10&page=1&sortBy=d,creationTimestamp'
+    url %= config['clusters'][0]['cluster']['server']
+    resp = await asyncify(requests.get)(url, auth=auth, verify=False)
+    assert resp.ok is True
+    data = resp.json()
+    assert len(data['podList']['pods'][0]['metrics']['cpuUsageHistory']) > 0
+    assert len(data['podList']['pods'][0]['metrics']['memoryUsageHistory']) > 0
+    with open(os.path.join(log_dir, 'dashboard.json'), 'w') as f:
+        yaml.dump(data, f, default_flow_style=False)
```
Cynerva: Should this be a `.yaml` file instead?

PR #76, opened 2017-07-21
Log e2e output

Cynerva:
This gives us access to e2e results produced by the e2e validator.


wwwtyro:
Yes, excellent, thanks @Cynerva!

PR #78, opened 2017-07-26
Feature/calico

ktsakalozos:


Cynerva:
@ktsakalozos Can you also update https://github.com/juju-solutions/kubernetes-jenkins/blob/master/charms/promote-all-charms-and-bundles.sh#L4 to include calico and kubernetes-calico? Other than that this looks great :+1:

ktsakalozos:
Done. Didn't think about that list. Sorry.

PR #82, opened 2017-08-14
Remove jujubox from gubernator

ktsakalozos:


Cynerva:
Looks like we're not using define-juju.sh anymore. Could you delete it?

ktsakalozos:
Done!

PR #83, opened 2017-08-23
Combine test_charms and test_snaps into test_cdk

Cynerva:
DO NOT MERGE. Checking this in will break our test-charms and test-snaps jobs. We should wait until after the release.

The purpose of this PR is to provide support for this test case:

* Test that master charms are working with snaps in the “next” track

Rather than add yet another test file, I think we can support this more cleanly by combining our existing tests in a more general way. This lets us use the following parameters to cover our main 3 test cases:

| Test Case | TEST_CHARM_CHANNEL | TEST_SNAP_CHANNEL |
| --------- | ------------------ | ----------------- |
| Test that master charms are working with stable snaps | edge | 1.7/stable |
| Test that stable charms work with latest upstream binaries in the current track | stable | 1.7/edge |
| Test that master charms are working with snaps in the “next” track | edge | 1.8/edge |


ktsakalozos:
LGTM looking forward to merging this

PR #87, opened 2017-08-31
Add canal bundle for testing

ktsakalozos:
DO NOT MERGE THIS YET.

Although the validations tests pass we cannot add this yet because we do not have anything in stable and we should also be granting access to everyone.


Cynerva:
I'm worried about the impact on our nightly test duration. The tests are already too long as it is.

That said, I'm good with merging this once the canal charms are released to stable. We can follow up later on how to address the long test duration.

PR #89, opened 2017-09-01
Disable elastic bundle build

Cynerva:
Our build-and-release-all-charms-and-bundles job currently fails with this:

```
+ bundle/bundle -o ./bundles/cdk-flannel-elastic -c edge k8s/cdk cni/flannel monitor/elastic
Processing fragment k8s/cdk...
Processing fragment cni/flannel...
Processing fragment monitor/elastic...
  File "bundle/bundle", line 204, in <module>
    main()
  File "bundle/bundle", line 175, in main
    bundle = version_bundle(bundle, args.channel)
  File "bundle/bundle", line 122, in version_bundle
    meta = [meta[k] for k in meta][0]
IndexError: list index out of range
```

This is happening because the newly promulgated elasticsearch charm hasn't been released to edge. We need a better fix, but for now, let's disable it so we get our builds over the weekend.

### DIFF
--- a/charms/build-and-release-bundles.sh
+++ b/charms/build-and-release-bundles.sh
@@ -14,13 +14,14 @@ release-bundle() {
 
 bundle/bundle -o ./bundles/cdk-flannel -c edge k8s/cdk cni/flannel
 bundle/bundle -o ./bundles/core-flannel -c edge k8s/core cni/flannel
-bundle/bundle -o ./bundles/cdk-flannel-elastic -c edge k8s/cdk cni/flannel monitor/elastic
+# Disabled elastic until the elasticsearch charm is back in edge
+# bundle/bundle -o ./bundles/cdk-flannel-elastic -c edge k8s/cdk cni/flannel monitor/elastic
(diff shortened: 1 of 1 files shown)


marcoceppi:
Couldn't we just put stable in the edge channel for that charm?

marcoceppi:
Furthermore, it's frustrating this exists, as the store should pick that
channel of the next stablest channel. Regardless of the fix here, the store
should be patched.

Cynerva:
> Couldn't we just put stable in the edge channel for that charm?

That'd be ideal. I don't have permissions on elasticsearch so this is just a temporary fix before we chase someone down next week.

> Furthermore, it's frustrating this exists, as the store should pick that
channel of the next stablest channel. Regardless of the fix here, the store
should be patched.

+1, yeah. This shouldn't really break us but it does.

marcoceppi:
I have access, I can do it in 2 hours time.

tvansteenburgh:
I just published it to edge/beta/candidate.

Cynerva:
Ah okay, didn't realize folks in our team had access. Thanks @marcoceppi and @tvansteenburgh

PR #93, opened 2017-10-03
Add bundle as cli so we can override it when on lxd

ktsakalozos:
On our CI we want to run the full set of tests on lxd, however, the calico & canal bundles will not deploy there. We need a way to override the set of bundles we run the tests against. With this PR for lxd we can do:

```
pytest -s test_cdk.py --bundle canonical-kubernetes
```

### DIFF
--- a/integration-tests/conftest.py
+++ b/integration-tests/conftest.py
@@ -21,3 +21,13 @@ def log_dir(request):
     )
     os.makedirs(path)
     return path
+
+
+def pytest_addoption(parser):
+    parser.addoption("--bundle", action="store", default=None,
+        help="bundle: specify the bundle under containers namespace")
+
+
+@pytest.fixture
+def bundle_override(request):
+    return request.config.getoption("--bundle")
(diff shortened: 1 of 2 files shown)


tvansteenburgh:
@ktsakalozos The commit message doesn't say why we need this; can you explain please?

Cynerva:
For the LXD case, I would consider adding something like this to the tests instead:

```
if cloud_is_lxd() and bundle in ['canonical-kubernetes-canal', 'kubernetes-calico']:
    pytest.skip('Skipping bundle %s on LXD' % bundle)
```

I'm very wary of adding a --bundle argument to work around this, since it's easy to forget an argument like that. I'd rather have our test infrastructure do the sane thing by default.

That said, being able to override the bundles for other reasons would probably be good. I'd prefer to do that as an environment variable, though, to be consistent with TEST_CHARM_CHANNEL and TEST_SNAP_CHANNEL.

ktsakalozos:
I like the env variable approach more than the `if cloud_is_lxd`.

I am going to put the card on lxd testing back to the backlog so we work on it when we get some time.

PR #95, opened 2017-10-04
Dump debug action output on test failures

Cynerva:
This should give us a lot more info to work with when debugging certain failures, e.g. the recent failure where kubedns never came up.


ktsakalozos:
LGTM +1

PR #98, opened 2017-10-12
Tests for RBAC

ktsakalozos:
Two tests on RBAC and a few small fixes
- select namespace fro where to deploy the test bundle so we can test bundles on our private namespaces.
- repeatable microbot test

### DIFF
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -305,10 +305,3 @@ async def run_bundletester(namespace, log_dir, channel='stable', snap_channel=No
             '-r', 'xml', '-o', output_file
         ]
         await asyncify(subprocess.check_call)(cmd)
-
-
-async def run_juju_ssh(unit, command):
-    ''' You can use this in case you need to do a unit.run("sudo ...") '''
-    cmd = ['juju', 'ssh', unit]
-    cmd += command.split()
-    await asyncify(subprocess.check_call)(cmd)
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -14,7 +14,7 @@ from async_generator import yield_
 from contextlib import contextmanager
 from juju.controller import Controller
 from juju.model import Model
-from juju.errors import JujuAPIError, JujuError
+from juju.errors import JujuError
 from subprocess import check_output, check_call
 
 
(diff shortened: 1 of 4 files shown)


integration-tests/utils.py
```diff
@@ -304,3 +305,10 @@ def asyncify(f):
             '-r', 'xml', '-o', output_file
         ]
         await asyncify(subprocess.check_call)(cmd)
+
+
+async def run_juju_ssh(unit, command):
+    ''' You can use this in case you need to do a unit.run("sudo ...") '''
+    cmd = ['juju', 'ssh', unit]
+    cmd += command.split()
+    await asyncify(subprocess.check_call)(cmd)
```
Cynerva: This function isn't used. Can we remove it?

integration-tests/validation.py
```diff
@@ -79,9 +80,61 @@ def validate_status_messages(model):
...
+    checks = 10
+    ready = False
+    while not ready:
+        checks -= 1
+        if await api_server_with_arg(model, arg):
+            return
+        else:
+            if checks <= 0:
+                assert False
+            asyncio.sleep(5)
```
Cynerva: should be `await asyncio.sleep(5)`

integration-tests/validation.py
```diff
@@ -79,9 +80,61 @@ def validate_status_messages(model):
...
+        assert action.status == 'completed'
+        raw_output = action.data['results']['Stdout']
+        return len(raw_output.splitlines()) == 1
+    return False
+
 async def validate_microbot(model):
     ''' Validate the microbot action '''
     unit = model.applications['kubernetes-worker'].units[0]
+    action = await unit.run_action('microbot', delete=True)
+    await action.wait()
```
Cynerva: YES! Thanks for this

Cynerva:
@ktsakalozos Looks like these tests need to be updated since the usage of authorization-mode changed. Can you confirm?

ktsakalozos:
You are right @Cynerva

integration-tests/validation.py
```diff
@@ -79,9 +80,61 @@ def validate_status_messages(model):
...
+    await wait_for_process(model, 'AlwaysAllow')
+    output = await worker.run(cmd)
+    assert output.status == 'completed'
+    assert "forbidden" not in output.data['results']['Stderr']
+
+
+async def validate_rbac_flag(model):
+    ''' Switch between auth modes and check the apiserver follows '''
+    master = model.applications['kubernetes-master']
+    await master.set_config({'authorization-mode': 'RBAC'})
```
Cynerva: Does this one need Node too?

ktsakalozos: I am just checking that the config is reflected on the kubeapi process. For this we can use any valid value. I have here only RBAC since RBAC+Node is indirectly tested by the previous test.

Cynerva:
Thanks @ktsakalozos, merging!

PR #99, opened 2017-10-17
Add validation for api-extra-args

Cynerva:
Don't merge. Initial testing with test_live_model looks good, but I still need to run this in test_cdk and make sure all is well.

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -234,7 +234,8 @@ async def validate_api_extra_args(model):
             action = await unit.run('pgrep -a kube-apiserver')
             assert action.status == 'completed'
             raw_output = action.data['results']['Stdout']
-            args = raw_output.partition(' ')[2].partition(' ')[2]
+            arg_string = raw_output.partition(' ')[2].partition(' ')[2]
+            args = {arg.strip() for arg in arg_string.split('--')[1:]}
             results.append(args)
         return results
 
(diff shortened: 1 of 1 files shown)


integration-tests/validation.py
```diff
@@ -222,5 +224,59 @@ def validate_status_messages(model):
...
+        all_args = await get_apiserver_args()
+        for args in all_args:
+            for expected_arg in expected_args:
+                if expected_arg not in args:
+                    await asyncio.sleep(1)
+                    continue
+        break
+    else:
+        raise TimeoutError('api-extra-args config did not propagate')
+
```
tvansteenburgh: Nitpick, but this block could be simplified (at least syntactically) to:

```python
while time.time() < deadline:
    all_args = await get_apiserver_args()
    if set(expected_args) <= set(all_args):
        break
    await asyncio.sleep(1)
else:
    raise TimeoutError('api-extra-args config did not propagate')
```

Cynerva: `all_args` is a list of full argument strings for each unit, something like:

```
[
    '--logtostderr=true --etcd-ca-file=/root/cdk/certs/ca.crt --etcd-cert-file=...',
    '--logtostderr=true --etcd-ca-file=/root/cdk/certs/ca.crt --etcd-cert-file=...'
]
```

To do what you're suggesting, we need to `.split()` the arg strings, and validate against all of them:

```
if all(set(expected_args) <= set(args.split()) for args in all_args):
    break
```

Guess I'll do that.

Also, `all_args` is confusingly named. I'mma rename it to `args_for_all_units` or something.

Cynerva:
Simplified the arg comparison logic using set comparison. Thanks @tvansteenburgh.

Turns out my original logic was broken anyway because the `continue` was captured by the innermost for loop. Derp.

Got a passing run with test_cdk, so if this looks good then feel free to merge.

PR #100, opened 2017-10-18
Feature: configurable list of bundles 

ktsakalozos:
Make list of bundles to test configurable from an conf variable. This allows us to test on lxd.

### DIFF
--- a/integration-tests/test_cdk.py
+++ b/integration-tests/test_cdk.py
@@ -7,8 +7,9 @@ from validation import validate_all
 namespace = os.environ.get('TEST_CHARM_NAMESPACE', 'containers')
 charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
 snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.8/stable')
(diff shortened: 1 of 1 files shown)


integration-tests/test_cdk.py
```diff
@@ -7,11 +7,8 @@
 namespace = os.environ.get('TEST_CHARM_NAMESPACE', 'containers')
 charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
 snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.8/stable')
-bundles = [
-    'canonical-kubernetes-canal',
-    'kubernetes-core',
-    # 'canonical-kubernetes',
-]
+bundles = os.environ.get('TEST_BUNDLES', 
+                         "'canonical-kubernetes-canal','kubernetes-core'").split(',')
```
tvansteenburgh: These strings are double-quoted. Also we should probably strip whitespace after the split, in case someone does:
```
TEST_BUNDLES="canonical-kubernetes-canal, kubernetes-core"
```

PR #101, opened 2017-10-23
Check for new release. Build snaps and release to edge and candidate.

ktsakalozos:
This PR adds build-and-release-snaps-on-new-release.sh that is meant to be triggered periodically from our CI. This script will scan for new stable releases and will build the snaps and push them to edge and candidate using the respective scripts.

The promote snaps script is patched to allow to actually promote the snaps instead of faking it.

### DIFF
--- a/snaps/build-and-release-snaps-on-new-release.sh
+++ b/snaps/build-and-release-snaps-on-new-release.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 #
 
+set -eux
+
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 # LAST_RELEASE_FILE keeps the last release we did.
 LAST_RELEASE_FILE="/var/tmp/last_k8s_patch_release"
@@ -14,7 +16,6 @@ function check_for_release {
     if [ "$LAST_RELEASED" != "$KUBE_VERSION" ]
     then
       echo "New release ($KUBE_VERSION) detected."
-      echo "$KUBE_VERSION" > $LAST_RELEASE_FILE
       trigger='yes'
     else
       echo "No new release detected. Latest release is $LAST_RELEASED."
@@ -36,12 +37,12 @@ then
   # Build the snaps and push to edge
   $scripts_path/build-and-release-k8s-snaps.sh
 
-  if [ "$?" == 0 ]
-  then
-    # Promote snaps from edge to candidate
-    version=${KUBE_VERSION:1:3}
-    export PROMOTE_FROM="$version/edge"
-    export PROMOTE_TO="$version/beta $version/candidate"
-    $scripts_path/promote-snaps.sh
-  fi
+  # Promote snaps from edge to candidate
+  version=${KUBE_VERSION:1:3}
+  export PROMOTE_FROM="$version/edge"
+  export PROMOTE_TO="$version/beta $version/candidate"
+  $scripts_path/promote-snaps.sh
+
+  # We are done with promoting the snaps. Lets mark the release.
+  echo "$KUBE_VERSION" > $LAST_RELEASE_FILE
 fi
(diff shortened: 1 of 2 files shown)


tvansteenburgh:
This is a good start but lacks a couple of things.

1. This only works for the latest minor release (1.8.x right now), but we need to be able to do this in older tracks too (1.7, 1.6, etc). Those tracks get patch releases too.
2. Looks like this will require the final promotion to stable (after our one-week buffer period) to be done manually. That may be fine at first but we'll want it automated soon or it will become too tedious to keep up with.

snaps/promote-snaps.sh
```diff
@@ -7,18 +7,29 @@ set -eu
...
   revisions="$(snapcraft revisions $snap | grep " ${PROMOTE_FROM}\*" | cut -d " " -f 1)"
   for rev in $revisions; do
     for target in $PROMOTE_TO; do
-      echo snapcraft release $snap $rev $target
+      if [ "${FAKE_PROMOTE}" = "yes" ]
+      then
+        echo snapcraft release $snap $rev $target
+      else
+        retry snapcraft release $snap $rev $target
+      fi
```
Cynerva: Is there any value in preserving the FAKE_PROMOTE=yes case? I'd rather just remove the `echo`s.

That was originally there just to play it safe when releasing to stable since I wasn't confident this script would do the right thing. We've used this script for at least four releases now and haven't seen any issues. I think we're good to just promote directly in all cases.

ktsakalozos: +1

snaps/build-and-release-snaps-on-new-release.sh
```diff
@@ -0,0 +1,48 @@
...
+  scripts_path=$(dirname "$0")
+  # Build the snaps and push to edge
+  $scripts_path/build-and-release-k8s-snaps.sh
+
+  if [ "$?" == 0 ]
+  then
+    # Promote snaps from edge to candidate
+    version=${KUBE_VERSION:1:3}
+    export PROMOTE_FROM="$version/edge"
+    export PROMOTE_TO="$version/candidate"
```
Cynerva: Can we promote to `$version/beta` too? I don't think we should ever have 1.x/beta be older than 1.x/candidate.

PROMOTE_TO takes a space separated list, so you can just shove it in there too.

snaps/build-and-release-snaps-on-new-release.sh
```diff
@@ -0,0 +1,47 @@
...
+function check_for_release {
+  trigger='no'
+  if [ -f $LAST_RELEASE_FILE ]
+  then
+    LAST_RELEASED=`cat $LAST_RELEASE_FILE`
+    if [ "$LAST_RELEASED" != "$KUBE_VERSION" ]
+    then
+      echo "New release ($KUBE_VERSION) detected."
+      echo "$KUBE_VERSION" > $LAST_RELEASE_FILE
+      trigger='yes'
```
Cynerva: What if build-and-release-k8s-snaps.sh crashes? e.g. snap store is down. This script will have updated the version in $LAST_RELEASE_FILE without that version ever being pushed or released. So we could end up skipping over a minor release if we're unlucky.

snaps/build-and-release-snaps-on-new-release.sh
```diff
@@ -0,0 +1,47 @@
...
+
+  if [ "$?" == 0 ]
+  then
+    # Promote snaps from edge to candidate
+    version=${KUBE_VERSION:1:3}
+    export PROMOTE_FROM="$version/edge"
+    export PROMOTE_TO="$version/beta $version/candidate"
+    $scripts_path/promote-snaps.sh
+  fi
+fi
```
Cynerva: If build-and-release-k8s-snaps.sh fails, this script will indicate success (exit code 0). I'd like to know if it fails.

Cynerva:
:+1: Merging. Make sure you follow up on @tvansteenburgh's points.

PR #102, opened 2017-10-24
Check for releases on all branches and release snaps

ktsakalozos:


marcoceppi:
Not a technical review, just saying I like where this is heading!

snaps/build-and-release-snaps-on-new-release.sh
```diff
@@ -3,12 +3,14 @@
 
 set -eux
 
-KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
-# LAST_RELEASE_FILE keeps the last release we did.
-LAST_RELEASE_FILE="/var/tmp/last_k8s_patch_release"
-
+# K8s versions we want to check for new releases
+declare -a VERSIONS=('1.6' '1.7' '1.8' '1.9' '2.0' 'latest');
```
Cynerva: I get the impression from [here](https://github.com/kubernetes/kubernetes/milestones/) that we will be getting 1.10, 1.11, ... before we get a 2.0.

This will probably cause problems when we slice the version with `${KUBE_VERSION:1:3}`, but that is not a problem specific to this PR; see [existing code that suffers from this as well](https://github.com/juju-solutions/kubernetes-jenkins/blob/master/snaps/build-and-release-k8s-snaps.sh#L25-L34). I will make a separate card to address this concern.

marcoceppi: Could we just read the refs from git? https://github.com/kubernetes/kubernetes/branches `release-X.Y`?

Cynerva:
Looks good, thanks @ktsakalozos

snaps/build-and-release-snaps-on-new-release.sh
ktsakalozos: @marcoceppi indeed this is how I initially approached this task. There are a few issues that will get us into trouble if we use the git tags. Normally the tags are created on the repo first and then the build process produces the kubernetes binaries. Our snaps need/repackage these binaries, so if we detect a new git tag we would still wait for the binaries to be uploaded. Another case where we would need to delay the snap build is when a tag is created pointing to a commit and then that tag is moved forward to include a bug fix (do not know if this has ever happened in k8s). Other minor annoyances, include tags that do not yield a release eg, beta-1. We would have to filter those out.

PR #103, opened 2017-10-24
Use CDK for e2e tests and error on gubernator errors

ktsakalozos:
A few fixes:
1. Gubernator jobs fail if e2e fail
2. Wait for the deployment to become ready will wait for all apps to get ready (not only kubernetes-master)
3. Use CDK and not core bundle


Cynerva:
:+1:

PR #106, opened 2017-10-26
Use stricter timeout guards in validate_api_extra_args

Cynerva:
This makes the timeouts in `validate_api_extra_args` work even if something within the loop hangs forever. (Example: https://github.com/juju/python-libjuju/pull/177)


ktsakalozos:
LGTM +1

PR #107, opened 2017-10-27
Build cdk addons from release branches

ktsakalozos:
With this PR we are cheking out the branch "release-X.Y" of cdk addons when we build against X.Y.Z k8s.

If the cdk-addons branch is not there that means we are building addons for the latest k8s and we default to head master.

### DIFF
--- a/snaps/build-and-release-k8s-snaps.sh
+++ b/snaps/build-and-release-k8s-snaps.sh
@@ -21,6 +21,7 @@ git clone https://github.com/juju-solutions/release.git --branch rye/snaps --dep
 rm -rf ./cdk-addons
 if git ls-remote --exit-code --heads https://github.com/juju-solutions/cdk-addons.git ${ADDONS_BRANCH_VERSION}
 then
+  echo "Getting cdk-addons from ${ADDONS_BRANCH_VERSION} branch."
   git clone https://github.com/juju-solutions/cdk-addons.git --branch ${ADDONS_BRANCH_VERSION} --depth 1
 else
   echo "Branch for ${ADDONS_BRANCH_VERSION} does not exist. Getting cdk-addons from master head."
--- a/snaps/build-and-release-k8s-snaps.sh
+++ b/snaps/build-and-release-k8s-snaps.sh
@@ -4,10 +4,12 @@ set -eux
 
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 KUBE_ARCH="amd64"
-ADDONS_BRANCH_VERSION="release-${KUBE_VERSION:1:3}"
+
+source utilities.sh
+MAIN_VERSION=$(get_major_minor $KUBE_VERSION)
+ADDONS_BRANCH_VERSION="release-${MAIN_VERSION}"
 
 source utils/retry.sh
-source utilities.sh
 
 rm -rf ./release
 git clone https://github.com/juju-solutions/release.git --branch rye/snaps --depth 1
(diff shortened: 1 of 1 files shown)


snaps/build-and-release-k8s-snaps.sh
```diff
@@ -17,7 +18,13 @@ git clone https://github.com/juju-solutions/release.git --branch rye/snaps --dep
 )
 
 rm -rf ./cdk-addons
-git clone https://github.com/juju-solutions/cdk-addons.git --depth 1
+if git ls-remote --exit-code --heads https://github.com/juju-solutions/cdk-addons.git ${ADDONS_BRANCH_VERSION}
+then
+  git clone https://github.com/juju-solutions/cdk-addons.git --branch ${ADDONS_BRANCH_VERSION} --depth 1
```
hyperbolic2346: Do we want to add an echo here as well to say that we found the branch and we are using it?

Cynerva:
I'm +1 to building from release-X.Y branches, but I'm not sure how I feel about falling back to master.

So if we create a release-1.6 branch, that gives us a way to fix the build for 1.6.12. But if we don't create release-1.7 and release-1.8, then we still have potential to break those by accident with work for 1.9 or whatever. I can easily see us forgetting to create those release branches.

If release-X.Y does not exist, I would prefer one of the following to happen:
1. Job aborts with an error.
2. Check out master and create the release-X.Y branch automatically. (this might be too problematic)
3. If building from master, mark the release as coming from master, e.g. snap version `1.9.0+master`

Perhaps some combination of 1 and 3 would be best: fail if there is no release-X.Y branch, unless we manually tell it to use master; if we do, mark the release with +master.

Having said all that, the change in this PR is an improvement over what we have today, and will solve our problem for 1.6.12, so I'm +1 to merging it as-is.

ktsakalozos:
Let me also add here the need to know exactly the commit of cdk-addons used for each k8s point release. For example, we would like to know that for k8s 1.6.12 we used the cdk-addons commit XXXXXXXX . That essentialy means that we would like to tag the commit XXXXXX with 1.6.12 tag. @wwwtyro let me know if I miss anythng

ktsakalozos:
@Cynerva what would be the impact of marking a snap with a +master? Does this show up anywhere after we push it to the right channel?

Cynerva:
Yeah, in e.g. `snap list`:

```
$ snap list
Name            Version     Rev   Developer       Notes
cdk-addons      1.8.2+test  x1                    -
```

Fairly certain we would see it in `snapcraft status cdk-addons` as well. I guess you still have to go looking for it, though, so it doesn't help with my concern all that much.

snaps/build-and-release-k8s-snaps.sh
```diff
@@ -4,6 +4,7 @@ set -eux
 
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 KUBE_ARCH="amd64"
+ADDONS_BRANCH_VERSION="release-${KUBE_VERSION:1:3}"
```
Cynerva: @ktsakalozos Now that https://github.com/juju-solutions/kubernetes-jenkins/pull/108 has landed, can you fix this to use `get_major_minor` as well?

ktsakalozos:
Have a look at: https://trello.com/c/3CaYhmLg/241-release-snaps-for-every-k8s-point-release

You will find a description of what we want to do in this cycle, a summary of what we have done already and what are the next steps. Feel free to comment/edit the document.

I suggest we merge this PR to unblock the build and clarify the up commit tasks.

PR #109, opened 2017-10-30
Increase validate_api_extra_args timeouts

ktsakalozos:
Often validate_api_extra_args test times out. Have a look at:
https://ci.kubernetes.juju.solutions/job/build-and-release-all-charms-and-bundles/515/testReport/junit/(root)/test_cdk/test_deploy_kubernetes_core_/
https://ci.kubernetes.juju.solutions/job/build-and-release-all-charms-and-bundles/516/testReport/junit/(root)/test_cdk/test_upgrade_canonical_kubernetes_canal_/
https://ci.kubernetes.juju.solutions/job/build-and-release-all-charms-and-bundles/517/testReport/junit/(root)/test_cdk/test_deploy_canonical_kubernetes_canal_/

Was able to repro by trying this specific test many times.

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -307,22 +307,22 @@ async def validate_api_extra_args(model):
         'enable-swagger-ui=false'
     }
 
-    with timeout_for_current_task(60):
+    with timeout_for_current_task(180):
         while True:
             args_per_unit = await get_apiserver_args()
             if all(expected_args <= args for args in args_per_unit):
                 break
-            await asyncio.sleep(1)
+            await asyncio.sleep(3)
 
     original_args_config = original_config['api-extra-args']['value']
     await app.set_config({'api-extra-args': original_args_config})
 
-    with timeout_for_current_task(60):
+    with timeout_for_current_task(180):
         while True:
             new_args = await get_apiserver_args()
             if new_args == original_args:
                 break
-            await asyncio.sleep(1)
+            await asyncio.sleep(3)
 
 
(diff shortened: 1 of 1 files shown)


Cynerva:
Thanks @ktsakalozos.

This is rather unfortunate, since we're basically admitting that configuring api_extra_args can take up to 3 minutes to actually take effect. That's pretty damn long for something as simple as restarting a service with different arguments.

tvansteenburgh:
"...admitting that configuring api_extra_args can take up to 3 minutes to actually take effect."

Why is this the case? config-changed should happen almost immediately; do we know why it takes so long to apply the args and restart?

Cynerva:
Here's a sample of logging output during the config-changed hook after setting api-extra-args: https://gist.github.com/Cynerva/6b900f748f9967a03a9d1a69c2988b1a

This one took 40 seconds total.

17 reactive handlers were run, most of which took a second or so, some a bit longer. The longest I see is `on_config_api_extra_args_change` which took ~4 seconds. However, I'd say a majority of time is spent on handlers that are totally unrelated to changing the args, e.g. [push_service_data](https://github.com/kubernetes/kubernetes/blob/c2a18cca6b79d4f3764d50f2c418b965e32864a2/cluster/juju/layers/kubernetes-master/reactive/kubernetes_master.py#L512).

There's also a ~12 second delay at the end of the hook, after the handlers have run, before the hook finishes. No idea what's happening there.

It seems to me that config-changed does a lot more work than it needs to.

PR #111, opened 2017-11-02
Adding test for extra_sans

hyperbolic2346:
Verifies the certificate returned from the api server and nginx(for load balancer) is updated as extra_sans is changed.


wwwtyro:
This looks good to me. Nicely done!

PR #117, opened 2017-11-06
Validate controller-manager-extra-args and scheduler-extra-args

Cynerva:
Don't merge - we should probably wait until after the release, in case we need to run tests again for whatever reason.

This adds validation for https://github.com/kubernetes/kubernetes/pull/55185.


ktsakalozos:
LGTM +1

ktsakalozos:
I think I will merge this now

PR #118, opened 2017-11-09
Create lxc juju clients to perform parallel execution

ktsakalozos:
We have a few scripts to create, decommission lxc juju clients as well as copying juju credentials to those clients in case we change them on the master. Have a look at the jobs in this Jenkins view: https://ci.kubernetes.juju.solutions/view/Maintenance/

In this PR there is also a script that would create a job invocation per cloud, bundle, test combination. This script is used in this jenkins job: https://ci.kubernetes.juju.solutions/view/Charms%20and%20Bundles/job/test-cdk-parallel/

### DIFF
--- a/integration-tests/test_cdk_parallel.groovy
+++ b/integration-tests/test_cdk_parallel.groovy
@@ -1,0 +1,70 @@
+
+def tests = []
+def clouds = []
+def bundles = []
+def run_bt = false
+
+stage('Input Validation') {
+    echo "CLOUD_AWS: "+CLOUD_AWS
+    echo "CLOUD_GCE: "+CLOUD_GCE
+    echo "CLOUD_LXD: "+CLOUD_LXD
+    echo "TEST_CHARM_CHANNEL: "+TEST_CHARM_CHANNEL
+    echo "TEST_SNAP_CHANNEL: "+TEST_SNAP_CHANNEL
+    echo "TEST_BUNDLES: "+TEST_BUNDLES
+    echo "TESTS_NAMES: "+TESTS_NAMES
+
+    // Fill bundles array
+    bundles = TEST_BUNDLES.replace(' ','').split(',')
+
+    // Fill tests array
+    tests = TESTS_NAMES.replace(' ','').split(',')
+
+    // Fill the clouds array
+    if (CLOUD_AWS == "true") { clouds+="jenkins-ci-aws" }
+    if (CLOUD_GCE == "true") { clouds+="jenkins-ci-gce" }
+    if (CLOUD_LXD == "true") { clouds+="jenkins-ci-lxd" }
+
+    if (tests.contains("test_bundletester")) {
+        echo "Bundletester tests reuqested"
+        run_bt = true
+    }
+}
+stage('Testing') {
+    def jobs = [:]
+
+    for (int t = 0; t < tests.size(); t++) {
+        def test = tests[t]
+        if (test == "test_bundletester") continue
+        for (int c = 0; c < clouds.size(); c++) {
+            def cloud = clouds[c]
+            for (int b = 0; b < bundles.size(); b++) {
+                def bundle = bundles[b]
+                def job_name = (cloud+bundle+test).replace('_','').replace('-','')
+                echo "Preparing job for "+test+" on "+bundle+" on "+cloud
+                jobs[job_name] = {
+                    stage(cloud+" "+bundle+" "+test) {
+                        echo "Running "+job_name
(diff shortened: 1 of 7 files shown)


integration-tests/test_cdk_parallel.groovy
```diff
@@ -0,0 +1,70 @@
+
```
Cynerva: Can we put this script somewhere outside of `integration-tests/`? Anywhere else is fine. Maybe `tests/` makes sense.

Everything in `integration-tests/` is runnable by any of us on our own laptops. That's true today and I'd like to keep it true. So Jenkins scripts gotta go elsewhere.

maintenance/copy-juju-to-lxc-node.sh
```diff
@@ -0,0 +1,17 @@
...
+container_exists $container
+
+RUN rm -rf /root/.local/share/juju/*
+RUN rm -rf /var/lib/jenkins/.local/share/juju/*
+PUSH ~/.local/share/juju/accounts.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/models.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/controllers.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/credentials.yaml /root/.local/share/juju/
+RUN mkdir -p /var/lib/jenkins/.local/share/juju
+PUSH ~/.local/share/juju/foo.json /var/lib/jenkins/.local/share/juju/
```
Cynerva: What's foo.json?

maintenance/provision-lxc-node.sh
```diff
@@ -0,0 +1,56 @@
...
+                 'git+https://github.com/juju-solutions/cloud-weather-report' \
+                 'git+https://github.com/juju/juju-crashdump'
+RUN pip3 install 'git+https://github.com/juju/python-libjuju' \
+                 'git+https://github.com/juju-solutions/matrix' \
+                 'git+https://github.com/juju/amulet'
+
+
+RUN pip2 install -U charm-tools pyopenssl
+RUN pip3 install -U pytest  pytest-asyncio asyncio_extras juju requests pyyaml kubernetes
+RUN snap install conjure-up --classic
```
Cynerva: A few concerns about dependency stuff:

1. Could we clone and run `integration-tests/install-deps.sh`? That may need some updating, but if we use it in our automation regularly then we can make sure it stays relevant.
2. How will it work to upgrade dependencies on all the containers? Sometimes we need a new release of python-libjuju or whatever to fix a problem in the tests. Will that be easy?
3. juju coming from both ppa:juju/stable and the conjure-up snap might be OK, but it's a little scary. I've had weird problems with that in the past.
4. charm-tools coming from two places is also scary
5. Maybe unneeded stuff: cloud-weather-report, matrix

maintenance/provision-lxc-node.sh
```diff
@@ -0,0 +1,56 @@
...
+
+RUN mkdir -p /srv/artifacts
+RUN mkdir -p /root/.local/share/juju
+
+PUSH ~/.local/share/juju/accounts.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/models.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/controllers.yaml /root/.local/share/juju/
+PUSH ~/.local/share/juju/credentials.yaml /root/.local/share/juju/ 
+RUN mkdir -p /var/lib/jenkins/.local/share/juju
+PUSH ~/.local/share/juju/foo.json /var/lib/jenkins/.local/share/juju/
```
Cynerva: Here's foo.json again. Is this a real thing?

ktsakalozos: Lol, yes I saw this today. It is needed by the google provider https://ci.kubernetes.juju.solutions/job/test-cdk/96/console Will digg into this and let you know more soon.

ktsakalozos: That json file includes the gce credentials as described here: https://jujucharms.com/docs/2.1/help-google. The file name is unfortunate.

integration-tests/test_cdk_parallel.groovy
ktsakalozos: Done with the move.

Also fixed the script to handle the lxdnode we have. Ideally we would have an lxd controller same as all the others so we would not need any extra care for lxd, but for now we need to send the job to the specific lxd slave.

Cynerva:
:+1: Thanks @ktsakalozos

PR #121, opened 2017-11-17
Use default bundles per controller

ktsakalozos:


integration-tests/utils.py
```diff
@@ -347,3 +347,13 @@ def asyncify(f):
     cloud = await controller.get_cloud()
     await controller.disconnect()
     return cloud == 'localhost'
+
+
+def default_bundles():
+    loop = asyncio.get_event_loop()
+    localhost = loop.run_until_complete(is_localhost())
+    loop.close()
```
Cynerva: This `loop.close()` seems unsafe. Have a look at this test script:

```
import asyncio

loop = asyncio.get_event_loop()
loop.run_until_complete(asyncio.sleep(1))
loop.close()

loop2 = asyncio.get_event_loop()
assert loop2 is loop
loop2.run_until_complete(asyncio.sleep(1))
```

Running this results in an exception:
```
  File "test.py", line 9, in <module>
    loop2.run_until_complete(asyncio.sleep(1))
  File "/usr/lib/python3.5/asyncio/base_events.py", line 364, in run_until_complete
    self._check_closed()
  File "/usr/lib/python3.5/asyncio/base_events.py", line 334, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

In other words, `asyncio.get_event_loop()` always returns the same loop (unless we set a new one with `asyncio.set_event_loop`). We probably shouldn't close it here.

ktsakalozos: pytest must be doing something fancy because running the test from this branch does not fail: https://github.com/juju-solutions/kubernetes-jenkins/tree/test/loop 

Either case I will leave the loop open, it makes no difference in my tests.

Cynerva:
Thanks @ktsakalozos 👍

PR #122, opened 2017-11-17
Skip HostPath test on lxd e2e tests

ktsakalozos:


Cynerva:
Ooo didn't realize we could skip it so easily. Nice! Thanks @ktsakalozos

PR #123, opened 2017-11-18
Updating tests to support new master staged upgrades

hyperbolic2346:
Test updates to test changes in PR https://github.com/kubernetes/kubernetes/pull/55990


ktsakalozos:
lgtm +1

Cynerva:
Does this work when TEST_SNAP_CHANNEL is unset and defaults to 1.8/stable?

hyperbolic2346:
It did in my testing, but I'm not sure if I would expect that behavior to
hold. It seems in juju's best interest to only inform on an actual change,
but currently it seems to update on any change.

PR #126, opened 2017-11-24
Refactor snap upgrade tests

ktsakalozos:
Noticed we were missing a corner case: calling upgrade_snaps using the channel already present.

We can merge this or hold back until we figure out why the build is red.

### DIFF
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -287,30 +287,38 @@ async def upgrade_charms(model, channel):
 
 
 @log_calls_async
-async def upgrade_snaps(model, channel, app_name, blocking):
-    app = model.applications.get(app_name)
-    if not app:
-        raise Exception('upgrade_snaps: Unable to find {0} application!'.format(app_name))
-    await app.set_config({'channel': channel})
+async def upgrade_snaps(model, channel):
+    for app_name, blocking in {'kubernetes-master': True, 'kubernetes-worker': True, 'kubernetes-e2e': False}.items():
+        app = model.applications.get(app_name)
+        # missing applications are simply not upgraded
+        if not app:
+            continue
 
-    if blocking:
-        for unit in app.units:
-            # wait for blocked status
-            deadline = time.time() + 180
-            while time.time() < deadline:
-                if (unit.workload_status == 'blocked' and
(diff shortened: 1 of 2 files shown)


integration-tests/utils.py
```diff
@@ -278,33 +278,63 @@ def asyncify(f):
         except JujuError as e:
             if "already running charm" not in str(e):
                 raise
-    await wait_for_ready(model)
 
+    # When we upgrade from a non-rolling master to a rolling upgrade master
+    # the master units will block waiting for the user to call the upgrade action.
```
Cynerva: Ouch. I think we probably do not want this behavior in the charm - it's forcing the user to walk through a manual upgrade that does nothing but clear charm state.

ktsakalozos: The cause of this behaviour is the call to the  any_file_changed method https://github.com/kubernetes/kubernetes/blob/master/cluster/juju/layers/kubernetes-master/reactive/kubernetes_master.py#L109 The first call any_file_changed will signal file changes.

ktsakalozos: Yes this does not look good. I wonder what would happen on the initial deployment of the charm. Would we be blocked "Needing" to upgrade?

ktsakalozos: The first setup is doing a "forced" installation of snaps so we are do not block in that case.

On every installation we should be storing the "fingerprint" (aka call any_file_changed without waiting) of the snap resources we are installing so the check any_file_changed is doing is correct even the first time. For the upgrade from non-rolling to rolling master would it be ok not to block waiting for the admin to trigger the upgrade action, @Cynerva ?

Cynerva: > For the upgrade from non-rolling to rolling master would it be ok not to block waiting for the admin to trigger the upgrade action, @Cynerva ?

Yeah, that should be fine. Seems worth it to provide a more consistent user experience: `juju upgrade-charm` does not result in a block for manual upgrade. Only changing the channel config or resources should result in that.

Cynerva:
@ktsakalozos this has a conflict now, likely with the logging stuff I added (sorry!)

ktsakalozos:
Done with the rebase.

PR #127, opened 2017-11-30
Add test for CDK with docker from xenial-proposed

Cynerva:
We keep having to test new versions of docker from xenial-proposed, so here's a quick test for it.


ktsakalozos:
LGTM +1

PR #128, opened 2017-12-05
Adding cloud environment variable

hyperbolic2346:
Tests work on jaas now that you can force the cloud.


ktsakalozos:
Code seems ok to me. I do not know what to pass in the TEST_CLOUD to test this (it might have been obvious if I were a JaaS user).

hyperbolic2346:
Any cloud suitable for juju such as 'aws'. This is just a pass-through to the juju add-model command and as such can be any cloud available in `juju clouds`.

ktsakalozos:
So for JaaS is aws in case the model is on aws?

hyperbolic2346:
Yes, that will result in juju add-model aws, which tells JaaS to create the model on an aws cluster. Since JaaS is able to create models on different clouds you are required to specify when you call add-model.

PR #140, opened 2018-02-07
Fix flakiness in validate_docker_logins

Cynerva:
This fixes some failures I've been hitting while running this locally. In particular, the `docker login` command fails since it sometimes runs before the registry is responsive.

I suspect I hit this locally, but not on Jenkins, because jenkins-ci-aws is still on Juju 2.2.0. The `unit.run` calls seem to return a lot quicker on Juju 2.3.2.

I hit a few other sporadic failures while working on this, too, hence the introduction and widespread use of `run_until_success`. Bleh.

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -501,27 +501,34 @@ async def validate_sans(model):
 
 @log_calls_async
 async def validate_docker_logins(model):
-    kubectl_cmd = '/snap/bin/kubectl --kubeconfig /root/cdk/kubeconfig'
+    # Choose a worker. He shall be our vessel.
+    app = model.applications['kubernetes-worker']
+    vessel = app.units[0]
 
-    @log_calls_async
-    async def check_output(unit, cmd):
-        action = await unit.run(cmd)
-        assert action.status == 'completed'
-        if action.data['results']['Code'] != '0':
-            log('Command failed: ' + cmd)
-            log('unit: ' + unit.entity_id)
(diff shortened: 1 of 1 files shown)


integration-tests/validation.py
```diff
@@ -501,27 +501,34 @@ def validate_status_messages(model):
...
-            log('Command failed: ' + cmd)
-            log('unit: ' + unit.entity_id)
-            log('code: ' + action.data['results']['Code'])
-            log('stdout: ' + action.data['results']['Stdout'])
-            log('stderr: ' + action.data['results']['Stderr'])
-            raise Exception('Command failed: ' + cmd)
-        assert action.data['results']['Code'] == '0'
-        return action.data['results']['Stdout']
+    async def run_until_success(cmd):
+        while True:
```
ktsakalozos: Forever is a big word shouldn't there be some long timeout

ktsakalozos:
I am merging this and if we see we get stuck into the run_until_success() often we can add a max_retries or timeout.

integration-tests/validation.py
Cynerva: Yeah, I was on the fence about this. I figured we could let the [global test timeout](https://github.com/juju-solutions/kubernetes-jenkins/blob/master/integration-tests/utils.py#L145) catch it, but I wouldn't be opposed to adding a timeout here too - no smaller than 10 min, though.

PR #142, opened 2018-02-22
validate_docker_logins: Use emptyDir volume to persist data

Cynerva:
**Heads up:** This is completely untested due to today's charm store outage.

This should _hopefully_ fix the failures we've been seeing with a timeout during `validate_docker_logins`.

The problem is that the `test-registry` container gets restarted sometimes, and since there is no volume mount for the registry data, it loses any images that were uploaded to it. Since the `ubuntu` image is lost, `test-registry-user` gets stuck in a Pending state.

The fix is to add an `emptyDir` volume to `test-registry`, mounted at `/var/lib/registry`. This will allow the data to survive container restarts.

My original assessment that the Pod was getting rescheduled to a different Node is wrong. Pods don't change nodes. With that in mind, I've removed the `nodeSelector` because it accomplishes nothing.


ktsakalozos:
LGTM +1, passes locally.

PR #146, opened 2018-02-28
Add proxymode=userpsace to lxd deployments

ktsakalozos:


integration-tests/utils.py
```diff
@@ -244,15 +244,20 @@ def all_units_ready(model):
...
-                data = yaml.load(f)
             for app in ['kubernetes-master', 'kubernetes-worker']:
                 options = data['services'][app].setdefault('options', {})
                 options['channel'] = snap_channel
-            with open(data_path, 'w') as f:
-                yaml.dump(data, f)
+        if await is_localhost():
+            if 'options' not in data['services']['kubernetes-worker']:
+                data['services'][app].setdefault('options', {})
+            options = data['services'][app]['options']
```
Cynerva: Where does `app` come from? It looks like it comes from the `for` loop above, which is confusing, to say the least.

Cynerva:
Is this blocking testing for the release? I would normally propose that we fix this in the charms or LXD profile instead, but I don't want to hold up a release for that.

integration-tests/utils.py
ktsakalozos: You are right, it's kubernetes-worker

ktsakalozos:
Still running a couple of tests on lxd. I would not consider this blocking since we can ignore it for the release. This issue seems to be first reported https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/286 some time ago. I am +1 for fixing this in a more generic place but I'd rather not be in the charm since this is controller specific.

Cynerva:
Ok. I would request that we leave this out of the integration tests, then. If LXD is broken then the tests should reflect it.

PR #152, opened 2018-03-31
install charm-tools with snapd

kwmonroe:
Replace the `charm-tools` deb with a `charm` snap.  The former brings in a bunch of junk.  The latter does too, but it's in a giant black box, so it bothers me less.

Also the `charm-tools` deb is no longer maintained/updated/recommended.

### DIFF
--- a/integration-tests/install-deps.sh
+++ b/integration-tests/install-deps.sh
@@ -5,8 +5,9 @@ set -eux
 # Can be run again to upgrade dependencies.
 
 sudo apt update -yq
-sudo apt install -y unzip python3-pip python-pip squashfuse snapd charm-tools
+sudo apt install -y unzip python3-pip python-pip squashfuse snapd
 sudo snap install juju --classic
+sudo snap install charm
 sudo snap install conjure-up --classic
 sudo pip2 install 'git+https://github.com/juju/juju-crashdump'
 sudo pip2 install -U pyopenssl bundletester


ktsakalozos:
I vaguely remember that charm proof was not properly working without the charm-tools cumming from apt. I might be wrong or this might be fixed by now. The charms snap does include the charm proof. Just mentioning this here in case we see any charm proof errors. I am merging this now and please go ahead and recreate the Jenkins slaves with this set of packages. Thanks.

ktsakalozos:
Reverting this because snaps cannot access the jenkins workspace

kwmonroe:
@ktsakalozos boo!  Thanks for letting me know why we need the deb, though i really dislike it -- it brings juju debs and a slew of dependencies onto the master.

The issue seems to be because of snap confinement.  The 'charm' snap is strictly confined and works with the home interface.  Unfortunately, the jenkins $HOME is not under /home, so the apparmor profile that enables access to /home has no effect.

I did some quick tests to try and get /mnt/srv/jenkins or /var/lib/jenkins respected as a HOMEDIR in apparmor (including bind mounting /home), but I wasn't successful. Here's more info for future reference:

https://forum.snapcraft.io/t/how-can-i-use-snap-when-i-dont-use-home-user/3352/15

Alas, this can be a problem for another day.  Until then, polluted filesystems it is.

PR #155, opened 2018-04-02
dont fail in wait_for_ready if priming cmd returns 0 matches

kwmonroe:
Example failure:
https://ci.kubernetes.juju.solutions/job/e2e-gubernator-gce/336/

This happened because the `grep` that sets `lines` failed to find any matches, which is fine, but it leaves a non-zero $? that will cause any `set -e` scripts to fail immediately.  Fix it with an `|| true`.


hyperbolic2346:
LGTM

PR #156, opened 2018-04-02
Updating validation to support older dashboard urls

hyperbolic2346:
Not sure if we want to merge this if we no longer support 1.7, but here's a fix.


kwmonroe:
this should fix our issue with 1.7:
https://github.com/juju-solutions/bundle-canonical-kubernetes/issues/529

Issue #151, opened 2018-03-29
models remain after cancelling job


kwmonroe:
Start a `test-cdk` job, let it start deploying stuff, then cancel it via the jenkins UI.

SSH to the jenkins master and you'll see the model is still there with `juju models -c jenkins-ci-<cloud>`.  I think we need to trap a few exit signals to make sure models are destroyed if the job is cancelled.

In `layer-cwr`, we trap the following and call a cleanup handler:
```
trap "$handler || true" EXIT SIGHUP SIGINT SIGTERM
```

Maybe something similar would work for us, probably around [temporary_model](https://github.com/juju-solutions/kubernetes-jenkins/blob/master/integration-tests/utils.py#L140), maybe with a tasteful `atexit`.

We could also just run a cleanup job every few hours looking for `cdk-build-upgrade-x` models that have not had a "Last connection" in X hours:

```
$ juju models -c jenkins-ci-aws
Controller: jenkins-ci-aws

Model       Cloud/Region   Status     Machines  Cores  Access  Last connection
controller  aws/us-east-1  available         1      2  admin   just now
cdk-build-upgrade-7961  aws/us-east-1  available         3     12  admin   2017-10-23
```

kwmonroe:
I also find the `build-and-release-*` jobs quite offensive.  They do things like this:

```bash
  JUJU_CONTROLLER="jenkins-ci-${CLOUD}"
  juju switch ${JUJU_CONTROLLER}

  MODEL="build-and-release-etcd-${BUILD_NUMBER}"
  # Create a model just for this run of the tests.
  juju add-model ${MODEL}
  ...
  trap "juju destroy-model -y ${MODEL} || true" EXIT
```

The problem here is that the controller might change during execution of the job (either by an admin or some other job calling `juju switch`).  When this happens, `destroy-model` will be called on the wrong controller -- this will `fail || true`, so the job will still complete and no one but the accountant will be the wiser.

Fully qualify the trap with the controller name.

battlemidget:
This will be further addressed when all jobs are moved into pipelines

battlemidget:
All jobs are now in pipelines and cleanup is handled appropriately

PR #157, opened 2018-04-02
ensure correct model for gubernator tests

kwmonroe:
The gubernator tests rely on "juju switch" being in the right
model before processing other juju commands. This is fragile
because an admin or other job might switch the controller during
a test run.

Ensure all juju commands for gubernator tests specify the
'controller:model' option.

Note, this also involves a change in the e2e-gubernator-* jobs
in jenkins to export a new CONTROLLER envar before the test begins.

This partially addresses #151 .

### DIFF
--- a/tests/deploy-test-bundle.sh
+++ b/tests/deploy-test-bundle.sh
@@ -7,27 +7,32 @@ set -o xtrace  # Print the commands that are executed.
 
 echo "${0} started at `date`."
 
-# First argument is the model namme for this build.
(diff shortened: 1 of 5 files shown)


kwmonroe:
Latest runs for both AWS and GCE look good when executed from this branch:

https://ci.kubernetes.juju.solutions/job/e2e-gubernator-aws/493/
https://ci.kubernetes.juju.solutions/job/e2e-gubernator-gce/339/

PR #169, opened 2018-04-05
Wait for registry pods to terminate before removing the mounted paths

ktsakalozos:
There is a race condition where we delete the directory with the registry secret before the pod has the opportunity to unmounted. This causes the registry pod to stay in Terminating state forever.

Fixes https://github.com/juju-solutions/kubernetes-jenkins/issues/168


ktsakalozos:
Merging this to queue some more tests

PR #170, opened 2018-04-05
Add deadline to registry

ktsakalozos:


integration-tests/validation.py
```diff
@@ -737,9 +739,18 @@ def validate_status_messages(model):
     registry_port = registry_service['spec']['ports'][0]['nodePort']
     registry_url = 'localhost:%s' % registry_port
 
+    log('Waiting for service to come up...')
+    while True:
```
hyperbolic2346: why not use the verify_ready function in this file to check for the service and pod?

kwmonroe: doh!  that's a good catch.  i suspect @ktsakalozos didn't realize you added this function recently.

ktsakalozos: The verify_ready solution would look like:
```
    deadline = time.time() + 600
    while time.time() < deadline:
        if await verify_ready(unit, 'po', ['test-registry'], '-n defautl') and \
               await verify_ready(unit, 'po', ['test-registry'], '-n defautl'):
            break
        await asyncio.sleep(5)
    else:
        raise TimeoutError('Unable to find test-registry before timeout')
```

This is almost the same amount of lines so I do not gain anything from there. I decided to go with the `while True` (although I do not like it) because this is how the rest of the validation function is. +1 for refactoring the tests to be homogeneous.

PR #175, opened 2018-04-18
Updating integration tests to take an upgrade channel

hyperbolic2346:
Facilitates more comprehensive upgrade testing.

### DIFF
--- a/integration-tests/test_cdk.py
+++ b/integration-tests/test_cdk.py
@@ -5,12 +5,12 @@ from utils import juju_deploy, run_bundletester, default_bundles
 from validation import validate_all
 
 namespace = os.environ.get('TEST_CHARM_NAMESPACE', 'containers')
-charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
-snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.9/stable')
+test_charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
+test_snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.9/stable')
 test_cloud = os.environ.get('TEST_CLOUD', '')
 bundles_csv = os.environ.get('TEST_BUNDLES', default_bundles())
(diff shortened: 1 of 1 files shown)


integration-tests/test_cdk.py
```diff
@@ -9,6 +9,8 @@
 snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.9/stable')
 test_cloud = os.environ.get('TEST_CLOUD', '')
 bundles_csv = os.environ.get('TEST_BUNDLES', default_bundles())
+upgrade_snap_channel = os.environ.get('UPGRADE_SNAP_CHANNEL', '1.10/stable')
```
kwmonroe: Any reason why `1.10/stable` vs just `stable`?  The former means we'll have to remember to update this periodically.  The latter keeps current functionality.

integration-tests/test_cdk.py
```diff
@@ -5,22 +5,25 @@
...
 
 namespace = os.environ.get('TEST_CHARM_NAMESPACE', 'containers')
-charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
-snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.9/stable')
+test_charm_channel = os.environ.get('TEST_CHARM_CHANNEL', 'stable')
+test_snap_channel = os.environ.get('TEST_SNAP_CHANNEL', '1.9/stable')
 test_cloud = os.environ.get('TEST_CLOUD', '')
 bundles_csv = os.environ.get('TEST_BUNDLES', default_bundles())
+upgrade_from_snap_channel = os.environ.get('UPGRADE_FROM_SNAP_CHANNEL', 'stable')
+upgrade_from_charm_channel = os.environ.get('UPGRADE_FROM_CHARM_CHANNEL', 'edge')
```
kwmonroe: So close!!! You want `stable` here too, because that's what we want when doing the initial deploy during test_upgrade.  Rather moot, i know, because this will be overridden by the job, but let's do it right anyway.

PR #181, opened 2018-05-21
Support EKS snaps

kwmonroe:
For EKS, we want a subset of k8s snaps locked to the EKS k8s-master version.  Handle this with a new jenkins job that builds and releases "kubectl kubelet kube-proxy kubernetes-test" to a special eks channel.

Also update our promote-snaps job so we can specify a snap list (useful when we want to promote the eks subset).

### DIFF
--- a/snaps/build-and-release-aws-eks-snaps.sh
+++ b/snaps/build-and-release-aws-eks-snaps.sh
@@ -17,15 +17,11 @@ rm -rf ./release
 git clone https://github.com/juju-solutions/release.git --branch rye/snaps --depth 1
 (cd release/snap
   make KUBE_VERSION=$KUBE_VERSION KUBE_ARCH="$KUBE_ARCH" \
-    targets="kubectl kubelet kube-proxy"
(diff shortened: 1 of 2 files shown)


snaps/build-and-release-aws-eks-snaps.sh
```diff
@@ -0,0 +1,31 @@
...
+
+source utils/retry.sh
+
+rm -rf ./release
+git clone https://github.com/juju-solutions/release.git --branch rye/snaps --depth 1
+(cd release/snap
+  make KUBE_VERSION=$KUBE_VERSION KUBE_ARCH="$KUBE_ARCH" \
+    targets="kubectl kubelet kube-proxy"
+  make KUBE_VERSION=$KUBE_VERSION KUBE_ARCH="amd64" \
+    targets="kubernetes-test"
```
hyperbolic2346: If this is amd64 only and we want to prevent building it for other platforms, do we need to worry about not being able to run it? Should this be an error until we get kubernetes-test on other platforms or should we just not include it?

PR #187, opened 2018-06-04
Adding ppa needed for amulet

hyperbolic2346:


integration-tests/install-deps.sh
```diff
@@ -4,6 +4,7 @@ set -eux
 # Installs dependencies needed for upgrade tests
 # Can be run again to upgrade dependencies.
 
+sudo apt-add-repository ppa:juju/stable
```
kwmonroe: Pass a `-y` here, lest this script will hang waiting for input.

hyperbolic2346:
This should actually be in the test for the charms. Going to add it there instead.

hyperbolic2346:
Nevermind, this will be fine. :)

PR #189, opened 2018-06-05
If a charm has no resources, the short format prints 'No resources found.'

hyperbolic2346:
This caused use to think there were 3 resources: 'No', 'resources', 'found.'. This is sub-optimal so I am grepping it away.


hyperbolic2346:
https://github.com/juju-solutions/kubernetes-jenkins/issues/190

PR #195, opened 2018-06-08
support building/releasing snaps with a -suffix; make sure we're the right user

kwmonroe:
Build EKS snaps just like our others, but tack on an -eks suffix.  SNAP_SUFFIX is now supported by the build scripts (https://github.com/juju-solutions/release/pull/21), so adjust the eks job to use it.

Also double check that we are the right user before doing snapcraft release (eks snaps are managed by the cpc team).

### DIFF
--- a/snaps/build-and-release-aws-eks-snaps.sh
+++ b/snaps/build-and-release-aws-eks-snaps.sh
@@ -7,6 +7,9 @@ set -eux
 EKS_RELEASE="${EKS_RELEASE:-edge}"
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 KUBE_ARCH="amd64"
+
+# This suffix is used by the snap build script when we run make targets:
+#  https://github.com/juju-solutions/release/blob/rye/snaps/snap/build-scripts/build
 export SNAP_SUFFIX="eks"
 
 source utilities.sh
--- a/snaps/build-and-release-aws-eks-snaps.sh
+++ b/snaps/build-and-release-aws-eks-snaps.sh
@@ -4,13 +4,12 @@
 # This script will build and release snaps for AWS EKS.
 set -eux
 
-EKS_RELEASE="${EKS_RELEASE:-edge/eks.0}"
+EKS_RELEASE="${EKS_RELEASE:-edge}"
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 KUBE_ARCH="amd64"
+export SNAP_SUFFIX="eks"
 
 source utilities.sh
-MAIN_VERSION=$(get_major_minor $KUBE_VERSION)
-
 source utils/retry.sh
 
 rm -rf ./release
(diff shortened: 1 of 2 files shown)


kwmonroe:
I also updated the `promote-snaps.sh` script to ensure we're the right user if we're pushing snap variants.  The `build-and-release-x` job does this with a snapcraft login, but the `promote-snaps` job does not.

We could push the -variant logic from the script into the jenkins job itself, but i opted to handle this in the script to keep the job simpler.

snaps/promote-snaps.sh
```diff
@@ -6,19 +6,39 @@ set -eu
...
-GH_USER="${GH_USER:-}"
-GH_TOKEN="${GH_TOKEN:-}"
+# Set a flag if we're building the EKS variants
+if [[ "${SNAPS}" =~ "-eks" ]]; then
+  build_eks=true
+fi
+
+# Ensure we are the correct user
+if [ "$build_eks" = true ]; then
+  cp -a /var/lib/jenkins/.config/snapcraft/snapcraft-cpc.cfg \
```
kwmonroe: Meh, i don't like this anymore.  This opens us up for problems with concurrent jobs trying to twiddle the filesystem at the same time.

kwmonroe:
promote job has been updated with appropriate snapcraft login bits in the Build section:

https://ci.kubernetes.juju.solutions/view/Snaps/job/promote-snaps/

snaps/build-and-release-aws-eks-snaps.sh
```diff
@@ -20,8 +19,13 @@ git clone https://github.com/juju-solutions/release.git --branch rye/snaps --dep
     targets="kubectl kubelet kube-proxy kubernetes-test"
 )
 
-for app in kubectl kubelet kube-proxy kubernetes-test; do
+# Ensure we are the correct user (jenkins job does a 'snapcraft login $token')
+if ! $(snapcraft whoami | grep -q canonical-cloud-snaps); then
```
hyperbolic2346: good safeguard here

snaps/build-and-release-aws-eks-snaps.sh
```diff
@@ -4,13 +4,12 @@
 # This script will build and release snaps for AWS EKS.
 set -eux
 
-EKS_RELEASE="${EKS_RELEASE:-edge/eks.0}"
+EKS_RELEASE="${EKS_RELEASE:-edge}"
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
 KUBE_ARCH="amd64"
+export SNAP_SUFFIX="eks"
```
hyperbolic2346: unused?

kwmonroe: Nah, it's used -- just not here.  We need it in the env to be picked up by the build script here:

https://github.com/juju-solutions/release/blob/rye/snaps/snap/build-scripts/build#L46

I'll add a comment to that effect.

hyperbolic2346:
I'm thinking we should wait until the next release just to keep from causing trouble, what do you think @kwmonroe?

PR #196, opened 2018-06-08
Move github branch creation to promotion of stable.

hyperbolic2346:
The original plan of always working in master and creating branches for versions when a new version went stable falls over when you're working on multiple releases at a time. In our example, we have 1.10, 1.11, and 1.12. 1.10 is stable, 1.11 is in beta, and 1.12 is alpha. We need changes for 1.11, but don't want to break 1.10. To prevent breaking it, the decision was made to create branches for a release once the release becomes stable, so now it is in promote. Note that in our current setup, we need to merge changes
into the 1.10 branch if the changes are for both master(future) and the current(1.10) release. Also note that we can't make 1.12 unique changes until 1.11 hits stable, which is a limitation we're currently ok with here for simplicity.

### DIFF
--- a/snaps/build-and-release-k8s-snaps.sh
+++ b/snaps/build-and-release-k8s-snaps.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 #
 
-# This script appart from builgin and releasing the k8s snaps will also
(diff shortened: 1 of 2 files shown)


Cynerva:
LGTM :+1:

PR #199, opened 2018-06-15
Transfer .go-cookies to lxc-slaved

ktsakalozos:


ktsakalozos:
@johnsca is this the right way to handle go-cookies?

johnsca:
The `.go-cookies` file should allow authentication but it's an older format and I think it might expire.  You would probably be better using `~/.local/share/juju/cookies/<controller>.json` and `~/.local/share/juju/store-usso-token`, both of which should allow regenerating of `~/.go-cookies` as needed.

ktsakalozos:
When does .go-cookies get created? Without this "copy" operation I get this: https://ci.kubernetes.juju.solutions/job/test-deploy/71/console

johnsca:
If you're hitting that error then it means you don't have the `~/.local/share/juju/cookies/<controller>.json` file (see [`cookies_for_controller`](https://github.com/juju/python-libjuju/blob/master/juju/client/jujudata.py#L211-L219)).  If you can copy that file instead, you shouldn't need the `.go-cookies` file.  The `.go-cookies` file is the older format / location; it should work but I think using the per-controller cookie file is preferred.  But if it's easier to use that file rather than the per-controller file, it probably doesn't matter all that much.

ktsakalozos:
Thanks @johnsca I would rather stick with the single file because controllers may come and go as we extend our CI.

ktsakalozos:
Merging this. It is already tested.

PR #205, opened 2018-06-26
Update resources scripts

ktsakalozos:
Do not merge this yet. Lets wait for after the 1.10 release.

Updates the resources scripts because some paths have changes and also we do not build k8s enymore, we have k8s in snaps.


resources/jenkins-repackage-resources.sh
```diff
@@ -24,15 +24,13 @@ mkdir -p ${TEMPORARY_DIRECTORY}
 resources/repackage-easyrsa.sh
 
 export OS=$(get_os)
-export ARCHITECTURES=${ARCHITECTURES:-"amd64"}  #"amd64 arm arm64 ppc64le"
+export ARCHITECTURES=${ARCHITECTURES:-"amd64 arm64"}  #"amd64 arm arm64 ppc64le"
```
Cynerva: Why no s390x here? Is upstream not shipping s390x packages?

ktsakalozos: Indeed there is no s390x for etcd: https://github.com/coreos/etcd/releases needed for flannel here: https://github.com/juju-solutions/kubernetes-jenkins/blob/master/resources/repackage-flannel.sh#L32

PR #206, opened 2018-06-26
Make build-and-release-k8s-snaps.sh runnable on s390x

Cynerva:
@ktsakalozos I made some changes to make it possible to run this script on s390x. Do you think this is worth merging in?

1. Use `dpkg --print-architecture` instead of hardcoding to amd64
2. Remove the special cases that hardcode kube-controller-manager and kubernetes-test to amd64. Those special cases are very outdated and no longer relevant.
3. Add `SKIP_RELEASE_TAG` to make it possible to run this without GH_USER and GH_TOKEN. There is probably a better way to handle this but the tagging doesn't currently account for multiple architectures. I took the easy way out for now.
4. Remove `MAIN_VERSION` (which is currently passed in from Jenkins) and just use `VERSION` instead since they seem to be identical. I can't think of a case where they'd ever be different.

Of course, running this on our s390x machine with these changes is still a bit unwieldy:
```
HTTP_PROXY=http://squid.internal:3128 HTTPS_PROXY=https://squid.internal:3128 NO_PROXY=snapcraft.io,ubuntu.com SKIP_RELEASE_TAG=true KUBE_VERSION=v1.10.5 ./snaps/build-and-release-k8s-snaps.sh
```

Cc @kwmonroe since you've got arm64 on your plate

### DIFF
--- a/snaps/build-and-release-k8s-snaps.sh
+++ b/snaps/build-and-release-k8s-snaps.sh
@@ -10,7 +10,8 @@
 set -eux
 
 KUBE_VERSION="${KUBE_VERSION:-$(curl -L https://dl.k8s.io/release/stable.txt)}"
-KUBE_ARCH="amd64"
+KUBE_ARCH="$(dpkg --print-architecture)"
+SKIP_RELEASE_TAG="${SKIP_RELEASE_TAG:-false}"
 
 source utilities.sh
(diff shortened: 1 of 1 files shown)


ktsakalozos:
LGTM +1

PR #207, opened 2018-06-27
Fixes for running our tests on lxd

ktsakalozos:
A couple of fixes for running our tests on lxd.

- we need a special proxy setting.
- removed the f from `pgrep -af` because it was picking up processes that were not the services we were interested in.

### DIFF
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -261,6 +261,14 @@ async def juju_deploy(model, namespace, bundle, channel='stable', snap_channel=N
                 yaml.dump(data, f)
         await model.deploy(bundle_dir)
     await wait_for_ready(model)
+    # In localhost we need proxy-extra-args="proxy-mode=userspace"
+    if is_localhost():
+        workers = model.applications['kubernetes-worker']
+        new_config={
+            'proxy-extra-args': 'proxy-mode=userspace'
+        }
+        await workers.set_config(new_config)
+        await wait_for_ready(model)
 
 
 def asyncify(f):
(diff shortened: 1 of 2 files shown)


integration-tests/utils.py
```diff
@@ -261,6 +261,14 @@ def all_units_ready(model):
                 yaml.dump(data, f)
         await model.deploy(bundle_dir)
     await wait_for_ready(model)
+    # In localhost we need proxy-extra-args="proxy-mode=userspace"
```
kwmonroe: Should we set this back to whatever the prev value was when we're done testing?

integration-tests/utils.py
```diff
@@ -261,6 +261,14 @@ def all_units_ready(model):
...
+    if is_localhost():
+        workers = model.applications['kubernetes-worker']
+        new_config={
+            'proxy-extra-args': 'proxy-mode=userspace'
+        }
+        await workers.set_config(new_config)
+        await wait_for_ready(model)
```
Cynerva: Can we set this config sooner? Right now it looks to me like we do this:
1. deploy
2. wait_for_ready
3. set_config proxy-mode=userspace
4. wait_for_ready

It seems to me like it would be safer, and save time, to set it immediately after deploy and before the first `wait_for_ready`. Thoughts?

integration-tests/utils.py
ktsakalozos: No we just leave it like this. lxd deployments need to have this option set https://github.com/juju-solutions/bundle-canonical-kubernetes/wiki/Deploying-on-LXD

integration-tests/utils.py
ktsakalozos: Done!

PR #210, opened 2018-07-03
Fix arg format in validate_extra_args

Cynerva:
Fix for `TimeoutError: Timed out after 600.000000 seconds` error that is occurring in validate_extra_args on CI.

I broke this test with https://github.com/juju-solutions/release/pull/26. This PR fixes the test.


Cynerva:
Merging to fix CI, someone please review.

PR #211, opened 2018-07-04
Fix bug: await the is_localhost

ktsakalozos:


ktsakalozos:
Self merging this obvious bug. I need to get some runs from the ci.

Cynerva:
Good catch. LGTM

PR #212, opened 2018-07-09
Inject proxies and skip tests based on architecture

ktsakalozos:
We should now be able to `pytest -s --no-print-logs ./integration-tests/test_live_model.py` as soon as we bootstrap a controller.

### DIFF
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -253,34 +253,38 @@ async def juju_deploy(model, namespace, bundle, channel='stable', snap_channel=N
         data_path = os.path.join(bundle_dir, 'bundle.yaml')
         with open(data_path) as f:
             data = yaml.load(f)
-            if os.getenv('http_proxy') and os.getenv('https_proxy'):
-                # Lets inject proxy settings
-                http_proxy = os.environ.get('http_proxy')
-                https_proxy = os.environ.get('https_proxy')
-                snap_proxy = https_proxy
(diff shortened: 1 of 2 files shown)


integration-tests/utils.py
```diff
@@ -351,16 +364,32 @@ def asyncify(f):
...
             data = yaml.load(f)
-        if snap_channel:
-            for app in ['kubernetes-master', 'kubernetes-worker']:
-                options = data['services'][app].setdefault('options', {})
-                options['channel'] = snap_channel
-        data['services']['kubernetes-worker'].setdefault('options', {})['labels'] = 'mylabel=thebest'
-        yaml.Dumper.ignore_aliases = lambda *args: True
-        with open(data_path, 'w') as f:
-            yaml.dump(data, f, default_flow_style=False)
+            if os.getenv('http_proxy') and os.getenv('https_proxy'):
```
johnsca: This is a pretty big chunk of code to be copypasta'd.  Probably worth putting it in a function.  I mean, you made `arch` a function, but not all this?  ;)

johnsca:
LGTM other than the giant block of duplicated code.  Would really like to see that refactored before +1'ing.

integration-tests/utils.py
ktsakalozos: Easy!

ktsakalozos:
Thank you for the review @johnsca

PR #217, opened 2018-07-19
Do not use etcd 2.3 on s390x anymore

ktsakalozos:
Etcd 3.2 snap is now available on s390x. We should not patching the bundle yaml any more.


Cynerva:
LGTM!

PR #218, opened 2018-07-20
Use cdk-field-agent to collect debug info

Cynerva:
This makes it so the integration tests gather debug info using cdk-field-agent, instead of using built-in functions.

This is nice because cdk-field-agent gathers a lot of additional info that was missing from the integration logs, such as: charm config, model config, status history, and controller logs.

### DIFF
--- a/integration-tests/utils.py
+++ b/integration-tests/utils.py
@@ -3,6 +3,7 @@ import functools
 import json
 import os
 import random
+import requests
 import shutil
 import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


integration-tests/utils.py
```diff
@@ -20,50 +21,14 @@
...
-        await action.wait()
-        remote_path = action.data['results']['path']
-        filename = unit.name.replace('/', '_') + '.tar.gz'
-        local_path = os.path.join(result_dir, filename)
-        await scp_from(unit, remote_path, local_path)
-
-    coroutines = [dump_debug_action(unit) for unit in model.units.values() if unit]
-    await asyncio.wait(coroutines)
+def fetch_field_agent_and_run(log_dir):
+    url = 'https://raw.githubusercontent.com/juju-solutions/cdk-field-agent/master/collect.py'
```
Cynerva: This means the integration tests will need to be able to reach `raw.githubusercontent.com`. I think we should be able to access it via proxy in our internal testing machines, but I'm not sure. @ktsakalozos do you anticipate any problems with this requirement?

ktsakalozos:
LGTM +1 lets see how that goes!

PR #219, opened 2018-07-23
Add microk8s scripts

ktsakalozos:
With this PR we add a Jenkins job that would periodically run and trigger the LP builders for microk8s. This way microk8s snaps will stay up-to-date.

One secret needs to be available on Jenkins and two python libs on every slave so we can talk to the snapstore and LP.

This work needs more automation (branch and builders creation) but this is the first step where source branches and builders are created manually.

### DIFF
--- a/microk8s/build-and-release-on-new-upstream-release.py
+++ b/microk8s/build-and-release-on-new-upstream-release.py
@@ -11,6 +11,7 @@ tracks = ["latest", "1.10", "1.11", "1.12", "1.13", "1.14", "1.15"]
 
 
 def upstream_release(release):
+    """Return the latest stable k8s in the release series"""
     if release == "latest":
         release_url = "https://dl.k8s.io/release/stable.txt"
     else:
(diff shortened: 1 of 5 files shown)


microk8s/build-and-release-on-new-upstream-release.py
```diff
@@ -0,0 +1,82 @@
...
+                                     version='devel')
+
+    # get launchpad team data and ppa
+    snappydev = launchpad.people[configbag.people_name]
+
+    try:
+        # get snap
+        microk8s = launchpad.snaps.getByName(name=snap_name,
+                                               owner=snappydev)
+    except HTTPError as e:
```
johnsca: Why is this the only place where an `HTTPError` is caught?  Wouldn't other API calls potentially raise errors as well?  Do we even need to bother catching this one?

ktsakalozos: This is an exception which is somewhat expected. We may get this exception in case we do not have builders for this specific release. For example for 1.12 we may not create the builders immediately, so instead of raising the exception and failing the build we treat this error as an expected one. This is not 100% safe as we may be shadowing some true exceptions but I would expect any sporadic HTTP errors to be few and any failed builders builds to be triggered in the next cycle.

ktsakalozos:
Thank you @johnsca. We will have many opportunities to address any concerns we have. This is the first pass in automating the microk8s build&release process. We will have to further automate the source branching and builders creation.

PR #221, opened 2018-07-24
Correct the channel of the microk8s snap we are checking

ktsakalozos:
We have to look for a channel named for example 1.10/edge inside the 1.10 track.


ktsakalozos:
I do not think you could have seen this @johnsca without first doing some test runs :) Merging this now that we have the Jenkins job in place.

PR #239, opened 2018-09-10
Use git merge when pulling from upstream

Cynerva:
@battlemidget The rebase-every-2-hours thing hasn't been working out so well for us - when we have pull requests open against juju-solutions/kubernetes master, we frequently have to rebase them, otherwise the diffs are filled with nonsense (e.g. https://github.com/juju-solutions/kubernetes/pull/156/files) and we're not able to merge them.

It seems to be getting worse as we accumulate more commits that haven't made it to upstream, too. I don't think this is sustainable.

I liked @johnsca's suggestion from IRC the other day: merge when we pull, rebase when we push, try that and see how it works.

Can we try that?


battlemidget:
Yea, I'm no expert at git, so any help in this area is appreciated

PR #240, opened 2018-09-17
When validate_extra_args fails, log local vars

Cynerva:
This updates `validate_extra_args` to dump local vars when a timeout occurs. The output is pretty excessive, at almost 500 lines ([example](https://gist.github.com/Cynerva/f3529c1b1bda6e03179f3b80ec3154be)), but at least it'll show us what it's stuck on.

I didn't have much luck getting `jobs/integration/test_cdk.py` to run, so I made the same change to the older `integration-tests/validation.py` to test it there instead. Please let me know if you want me to back that out or try again with `jobs/integration/test_cdk.py` or whatever.

### DIFF
--- a/integration-tests/validation.py
+++ b/integration-tests/validation.py
@@ -10,6 +10,7 @@ from asyncio_extras import async_contextmanager
 from async_generator import yield_
 from datetime import datetime
 from logger import log, log_calls, log_calls_async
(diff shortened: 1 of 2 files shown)


battlemidget:
thanks! ill port it over to the other file

PR #244, opened 2018-10-12
Try to fix missing playbook

ktsakalozos:
@battlemidget I know this is not what you want. Since this file is missing I assume you were working on the new jenkins deployment. Not sure if this works, I am trying this out. @hyperbolic2346 Needs to run some ci tests.


battlemidget:
Yeah that works getting the tests migrated over to jenkaas so the playbook will change then, until then though feel free to move about the cabin

PR #265, opened 2018-11-13
Remove obsolete resource scripts

Cynerva:
Depends on:
* https://github.com/juju-solutions/kubernetes/pull/187
* https://github.com/juju-solutions/charm-flannel/pull/49

These old resource scripts don't get called anywhere, and once the above PRs are merged they will all have replacements, so I'm gonna suggest we just delete these.

Here's a breakdown.

| File | Comment
| --- | --- |
| build-cni.sh | Soon to be replaced by https://github.com/juju-solutions/kubernetes/pull/187 |
| build-etcd.sh | This resource no longer exist (charms install from snap store now) |
| build-flannel.sh | Soon to be replaced by https://github.com/juju-solutions/charm-flannel/pull/49 |
| build-kubernetes.sh | This resource no longer exists (charms install from snap store now) |
| jenkins-build-resources.sh | This just calls build-flannel.sh, which we're deleting |
| jenkins-repackage-resources.sh | This just calls repackage-flannel.sh, which we're deleting |
| repackage-flannel.sh | This fetches bins from upstreams, but we want to build our own instead |
| repackage-kubernetes.sh | This resource no longer exist (charms install from snap store now) |
| versions.sh | Redundant, defaults for versions are included in the scripts that use them |

The only one I did not delete is `repackage-easyrsa.sh`. That still needs a home - probably with layer-easyrsa.


Cynerva:
@battlemidget My intention here is to move the resource build scripts so that they live with the charm code, e.g. the script for building the flannel resource lives in the flannel charm repo. I do intend to create or modify Jenkins jobs to actually call them, but that will come next.

battlemidget:
Thanks @cynerva, if you catch the other prs merged before I do feel free to to merge this when ready

PR #268, opened 2018-11-16
Retry manually microk8s releases

ktsakalozos:
I've noticed the retry option does not work on our scripts:  https://jenkins.canonical.com/k8s/view/microk8s/job/release-microk8s-beta/78/consoleFull

There is this bug open for some time now: https://issues.jenkins-ci.org/browse/JENKINS-46354

So, this PR removes the retry but it schedules 4 builds throughout the day. When there is nothing to release the job takes 2 minutes, when we need to releases tunning the tests take about 1.5 hours.


battlemidget:
You could put retry on the actual stage and I know that works. Is that an option for you?

PR #271, opened 2018-11-21
Parallelise microk8s builds

ktsakalozos:
@battlemidget this is the second PR with work on microk8s multiarch. What i tried to do is to have the same tests running on amd64 and arm64. As long as the right agent is selected the tests will pick up the architecture and push to beta and stable the right snap revision.

Cannot say if this syntax works though, I need your help. Thanks.


ktsakalozos:
I do not like much that the code repeats. I am open to any suggestions on how we could handle this.

PR #281, opened 2018-12-04
Fix conflict between Keystone and Vault tests

johnsca:
Both use percona-cluster and there was a conflict with how they were deploying it.


hyperbolic2346:
I would like to see a test leave the state as close as possible to original. Any reason that percona isn't being cleaned up by vault?

johnsca:
@hyperbolic2346 The Vault test is actually run after the Keystone test, but it uses an overlay which I didn't realize gets applied before all the tests.  I was just discussing whether it would make more sense to move over to what you're doing.  This PR actually won't fix the issue since you do in fact clean up percona-cluster in the Keystone test.

PR #282, opened 2018-12-04
Refactor Vault test

johnsca:
Move away from using an overlay to explicitly deploy and cleanup dependencies. Removes additional conflict with Keystone test.


battlemidget:
We have a blocking test atm, so let's wait to merge this until that gets resolved.

PR #303, opened 2019-01-23
Add handling of oci-image resources to charms.py

knkski:
Adds logic that will pass in "--resource foo=bar" flags to "charm push", based on a new "upstream-source" attribute that defines a particular image to use.

Would like a :+1: from @johnsca and @wallyworld that these are the right resources for the currently-published charms in the charm store.

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -8,11 +8,14 @@ Usage:
 
   tox -e py36 -- python3 jobs/build-charms/charms.py --help
 """
+
+import os
+from glob import glob
+from pathlib import Path
+from pprint import pformat
+
 import click
 import sh
-import os
-from pathlib import Path
-from glob import glob
 import yaml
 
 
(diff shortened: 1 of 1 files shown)


Cynerva:
I don't think this will do what you want. The other entries in resource-spec.yaml are mapping filenames to resource names (you can how they're used [here](https://github.com/juju-solutions/kubernetes-jenkins/blob/2d7076dc07862f5300f1db736d6ae0718b41b8da/jobs/build-charms/charms.py#L101-L107)), not resource names to docker images.

battlemidget:
LGTM, waiting @johnsca and @wallyworld +1

jobs/build-charms/charms.py
```diff
@@ -8,11 +8,14 @@
...
   tox -e py36 -- python3 jobs/build-charms/charms.py --help
 """
-import click
-import sh
+
 import os
-from pathlib import Path
 from glob import glob
+from pathlib import Path
+from pprint import pformat
```
battlemidget: @knkski Can you remove this on next commit?

PR #323, opened 2019-03-11
Automated testing for Tigera EE and Calico

Cynerva:
Depends on https://github.com/battlemidget/juju-jenkins-pipeline/pull/2

This fixes various problems I see with the existing Calico jobs, and adds new jobs for testing Tigera Secure EE.

I did the following:
1. Rewrote `create-subnet-aws.sh` in Python (as `bootstrap_aws_single_subnet.py`). The script had broken yet again on AWS CLI output changes. Now it parses json output instead, which should alleviate this problem.
2. Rewrote `disable_source_dest_check.sh` in Python (as `disable_source_dest_check.py`), adding logic to only run on AWS and only when calico or tigera-secure-ee are present.
3. Fixed validation on Calico and Tigera Secure EE:
    1. Add `tigera-secure-ee` to the list of charms to run `validate_network_policy` on.
    2. Fixed `verify_ready` hanging when the string `"error"` is contained in otherwise healthy output.
    3. Fixed `validate_worker_master_scaling` to call `disable_source_dest_check.py` after adding units.
4. Fixed `validate.yaml` including a reference to the `kubernetes-calico` bundle. That bundle won't deploy cleanly with that job definition.
5. Fixed problems with the `validate-calico` job:
    1. 1.14 job was using the 1.13 overlay
    2. Juju controller was not bootstrapped with the VPC
    3. `juju-wait` was being run before `disable_source_dest_check.py`
6. Created `validate-tigera-secure-ee` job definition, which is nearly identical to `validate-calico`, but includes code for configuring the charm with the Tigera Secure EE license key and Tigera private registry credentials.


jobs/validate-tigera-secure-ee/Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+        }
+        success {
+            setPass()
+        }
+        always {
+            setEndTime()
+            collectDebug(params.controller,
+                         juju_model)
+
+            dir('jobs') {
```
battlemidget: Not a blocker, but it may be worth putting this in the `cleanup {}` stanza

Cynerva: Ah good catch, thanks! Opened a new PR for this: https://github.com/juju-solutions/kubernetes-jenkins/pull/325

PR #334, opened 2019-03-14
Use multipass for Kubeflow testing

knkski:
Uses multipass to spin up a VM that is then used for testing Kubeflow. This lets us choose specific version of snap dependencies, such as microk8s.


jobs/validate-kubeflow-microk8s/Jenkinsfile
```diff
@@ -60,15 +84,9 @@ pipeline {
         }
         always {
             setEndTime()
-            collectDebug(params.controller, juju_model)
-
         }
-        cleanup {
-            saveMeta()
```
battlemidget: Is there a way we can still capture some of this information? I utilize this for reporting failures etc

PR #342, opened 2019-03-23
Use xip.io address for Ambassador IP

knkski:
Since the routing is hostname-based.


knkski:
@battlemidget: I don't know of one offhand, this is the one that was recommended that I use in IRC. If there's one you'd prefer I use, I can definitely switch to that one.

battlemidget:
no worries, i just seem to remember xip.io being unreliable, but maybe that's changed

PR #362, opened 2019-04-03
use utils.scp_to() for localhost workaround

kwmonroe:
In a lxd cloud, `unit.scp_to` has trouble figuring out the address (often choosing the flannel interface vs eth0, for example). We work around this in other validation cases with `utils.scp_to`, which uses the unit name vs address in the scp invocation.

Looks like we missed this when coredns validation tests were included. Let's fix that post haste!


Cynerva:
Good catch. Thank you.

PR #363, opened 2019-04-04
Removing Docker specific validation

joedborg:
Because of the migration from docker layer to generic subordinate, we need to remove docker specific validation from the tests (and, eventually, move them to more specific tests).


jobs/integration/validation.py
```diff
@@ -785,209 +780,6 @@ def validate_status_messages(model):
             await asyncio.sleep(5)
 
 
-@log_calls_async
-async def validate_docker_logins(model):
```
Cynerva: The docker-logins config isn't so much a docker runtime thing, as it is a registry credential thing. What's happening to this config? What is replacing it?

If I have a private registry, and I need to provide a username and password for it, how do I do that when I'm using the new docker subordinate? How do I do it with the new containerd subordinate?

joedborg: @Cynerva this has moved to the subordinate charm itself: https://github.com/juju-solutions/charm-docker/pull/1/files.  It'll need adding to the containerd subordinate, but will have to be tested on a per-subordinate basis I guess?

jobs/integration/validation.py
```diff
@@ -454,12 +453,8 @@ def validate_status_messages(model):
     action = await workers.units[0].run('lspci -nnk')
     nvidia = True if action.results['Stdout'].lower().count("nvidia") > 0 else False
 
-    # See what the runtime is set to
-    config = await workers.get_config()
-    runtime = config['docker_runtime']['value']
-
     master_unit = model.applications['kubernetes-master'].units[0]
-    if not nvidia or runtime == 'apt' or runtime == 'upstream':
+    if not nvidia:
```
Cynerva: Before, on nvidia-enabled hardware, it was possible to opt out of nvidia support by setting docker_runtime=apt. Now it looks like there's no way to opt out. Will that be a problem?

LXD comes to mind. I've seen a lot of cases where LXD deployments detect nvidia hardware on the host, but are unable to use it, and that causes problems. The solution has been to set docker_runtime=apt. Do we have a new solution?

joedborg: @Cynerva This can still be done with the Docker subordinate (all the previous option remain, it's just about making these CDK tests generic).  If we think a `gpu_disable`-esque flag would be useful in the containerd subordinate, that's no issue.

jobs/integration/validation.py
Cynerva: Yeah, I don't think we want to lose this test. I'd prefer we update it to use the new config on docker if it's present.

Cynerva: I was just reminded that the plan is to remove engine-specific stuff here, but reintroduce this test elsewhere. That works for me.

jobs/integration/validation.py
Cynerva: > If we think a gpu_disable-esque flag would be useful in the containerd subordinate, that's no issue.

I guess I'd test containerd on LXD to see if it's needed. That's the only use case I'm aware of, at least.

joedborg: +1 I'll get that added to the charm config.

PR #365, opened 2019-04-04
Working Kubeflow testing on AWS

knkski:
Tested locally against Jenkins-created AWS infrastructure, and it works. Deploys with the AWS deploy script sitting over in `bundle-kubeflow`, to make sure it doesn't suffer from bitrot.


battlemidget:
ill sync the jobs in the morning

PR #388, opened 2019-04-25
Adding model-defaults for proxy

hyperbolic2346:

### DIFF
--- a/jobs/validate-hacluster/Jenkinsfile
+++ b/jobs/validate-hacluster/Jenkinsfile
@@ -28,6 +28,9 @@ pipeline {
                 sh "juju add-cloud scapestack /var/lib/jenkins/.scapestack/scapestack-cloud"
                 sh "juju add-credential scapestack -f /var/lib/jenkins/.scapestack/scapestack-creds"
                 sh "juju bootstrap scapestack ${params.controller} --debug"
+                sh "juju model-defaults http-proxy=http://squid.internal:3128"
+                sh "juju model-defaults https-proxy=http://squid.internal:3128"
+                sh "juju model-defaults no-proxy=localhost,10.96.0.0/17"
 
                 deployCDK(controller: params.controller,
                           cloud: params.cloud,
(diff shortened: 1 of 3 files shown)


jobs/validate-hacluster/Jenkinsfile
```diff
@@ -28,6 +28,9 @@ pipeline {
                 sh "juju add-cloud scapestack /var/lib/jenkins/.scapestack/scapestack-cloud"
                 sh "juju add-credential scapestack -f /var/lib/jenkins/.scapestack/scapestack-creds"
                 sh "juju bootstrap scapestack ${params.controller} --debug"
+                sh "juju model-defaults http-proxy=http://squid.internal:3128"
+                sh "juju model-defaults https-proxy=http://squid.internal:3128"
+                sh "juju model-defaults no-proxy=localhost,10.96.0.0/17"
```
Cynerva: I wouldn't recommend using these model configs. They get added to /etc/environment on every machine and it usually causes problems with etcdctl.

Proper proxy config is documented here: https://github.com/charmed-kubernetes/bundle/wiki/Proxy-configuration

In practice, I've almost never needed the juju-* proxy model configs. But you do need the snap_proxy, http_proxy, and https_proxy charm configs.

hyperbolic2346: So you would suggest config on the charms instead of using juju model config? The doc you linked suggests both.

hyperbolic2346: Added overlays, but interested if that was in addition as the document suggests or in place of the model-defaults. Sounds like you're in favor of removing the model-defaults.

Cynerva: The doc suggests setting these model configs: `juju-http-proxy`, `juju-https-proxy`, `juju-no-proxy`. In practice, I have never needed these, but they've also never hurt. We document them mostly because Juju says to use them. You're free to set them here too.

The doc does not suggest setting these model configs: `http-proxy`, `https-proxy`, `no-proxy`. That's for a very specific reason: they cause problems with etcdctl. I'd link you the context but it was an issue in the layer-etcd repo. We lost those. :(

Either way, you need the charm configs.

hyperbolic2346: ah, I missed the distinction between juju-http-proxy and http-proxy, sorry about that.

Cynerva: No worries. The doc didn't make it very clear.

Cynerva:
LGTM

PR #397, opened 2019-05-03
Feature: Release MicroK8s pre-releases

ktsakalozos:
---

- For Integration tests (test_cdk.py) please make those changes in `jobs/integration`
- For microk8s, those changes are in `jobs/build-microk8s`

### DIFF
--- a/jobs/microk8s/utils.py
+++ b/jobs/microk8s/utils.py
@@ -1,0 +1,15 @@
+import requests
+
+
+def upstream_release(release):
+    """Return the latest stable k8s in the release series"""
+    if release == "latest":
+        release_url = "https://dl.k8s.io/release/stable.txt"
+    else:
+        release_url = "https://dl.k8s.io/release/stable-{}.txt".format(release)
+
+    r = requests.get(release_url)
+    if r.status_code == 200:
+        return r.content.decode().strip()
+    else:
+        None
--- a/jobs/microk8s/utils.py
+++ b/jobs/microk8s/utils.py
@@ -1,5 +1,6 @@
 import requests
 import json
+import semver
 from snapstore import Microk8sSnap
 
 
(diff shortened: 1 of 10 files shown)


jobs/microk8s/utils.py
```diff
@@ -0,0 +1,213 @@
...
+    if arevision != brevision:
+        if arevision > brevision:
+            return 1
+        elif arevision < brevision:
+            return -1
+
+    # Since we reached this spot we have major, minor, patch and revision sections equal
+    # This should have been caught at the very beginning.
+    assert False
+
```
battlemidget: They use semver, you could use the semver library to handle this, ie https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/build-snaps/snaps-source.py#L46-L53

jobs/microk8s/utils.py
```diff
@@ -0,0 +1,213 @@
...
+    print(releases)
+
+    releases = get_latest_pre_release("1.14","rc")
+    print(releases)
+
+    releases = get_latest_pre_release("1.12","rc")
+    print(releases)
+
+    snap = Microk8sSnap("1.15", "edge", juju_controller="aws-us-east-1:default", juju_unit="ubuntu/0")
+    snap.build_and_release("1.15.0-alpha.2")
```
battlemidget: Not sure if you're interested, but i took a different approach, rather than having all these versions hardcoded, i just take the tag releases from upstream and make sure they are synced in the snap store

https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/build-snaps/snaps-source.py#L36-L77

Create branches for the ones we support and dont exist in launchpad and just have them built automatically. 

Also, now that I think about it, you could just use re-use some of the snaps I build from source and set your snapcraft to use stage-snaps, https://docs.snapcraft.io/snapcraft-parts-metadata (see `stage-snaps`)

jobs/microk8s/utils.py
ktsakalozos: Awesome, less code less bugs :)

jobs/microk8s/utils.py
ktsakalozos: This was misleading, sorry. This part of the code was only used for local testing. Removed.

battlemidget:
:tada:

PR #413, opened 2019-05-15
check for stable vs latest and only create addons branch for stable

kwmonroe:
Move the K8s version code higher so we know if we have a stable or a latest k8s version.  Only create the cdk-addons release branch on stable.


Cynerva:
LGTM, thanks @kwmonroe

PR #443, opened 2019-06-25
Fix Vault test not actually running

johnsca:
The condition to run the Vault test logic was missed in the refactor in PR #282 causing the Vault parts of the tests to never run. This changes it to drive the Vault test off the model name.

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -59,7 +59,7 @@ async def validate_all(model, log_dir):
     if cpu_arch not in ['s390x', 'arm64', 'aarch64']:
         await validate_keystone(model)
     assert_no_unit_errors(model)
-    if ('validate-vault' == model.name and
+    if ('validate-vault' == model.info.name and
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -59,7 +59,7 @@
     if cpu_arch not in ['s390x', 'arm64', 'aarch64']:
         await validate_keystone(model)
     assert_no_unit_errors(model)
-    if ('vault' in model.applications and
+    if ('validate-vault' == model.info.name and
```
Cynerva: Should be `in` not `==`, given that model name is generally `validate-vault-<uuid>`:
https://github.com/charmed-kubernetes/jenkins/blob/ceba5490a35ac806b237ed7fc247e846ad0cc0f7/jobs/validate-vault.yaml#L14-L15
https://github.com/charmed-kubernetes/jenkins/blob/ceba5490a35ac806b237ed7fc247e846ad0cc0f7/jobs/validate-vault/Jenkinsfile#L3

PR #444, opened 2019-06-25
Add canal empty resouce

joedborg:
Only merge in now if it doesn't effect the release.


Cynerva:
build-release-canal is going to fail until this gets merged (or we revert https://github.com/charmed-kubernetes/layer-canal/pull/34)

PR #447, opened 2019-07-02
Tests for juju-http(s)-proxy environment variables

VariableDeclared:
Added tests to check layer-docker, charm-docker, and charm-containerd for correct setting of HTTP(S) environment variables.


jobs/integration/test_juju_https_proxy.py
```diff
@@ -0,0 +1,191 @@
+import asyncio
+import pytest
+import re
+import os
+import time
+from .base import (
+    UseModel,
+    _juju_wait
+)
```
battlemidget: You dont need UseModel anymore and _juju_wait has moved to utils

jobs/integration/test_juju_https_proxy_legacy.py
```diff
@@ -0,0 +1,156 @@
+import asyncio
+import pytest
+import re
+import os
+import time
+from .base import (
+    UseModel,
+    _juju_wait
+)
```
battlemidget: same as above

PR #456, opened 2019-07-20
Fix end-to-end http(s) proxy tests

VariableDeclared:
Added fixture and removed unnessary tests for legacy functions.

Also cleaned up tests.


battlemidget:
Looking much much better

VariableDeclared:
@battlemidget cheers for the feedback. Should have this finished before midday.

PR #457, opened 2019-07-26
Adding container runtime and add option to juju wait

joedborg:


joedborg:
[Tests for container runtime subordinates](https://trello.com/c/LkyaiAXE/1218-tests-for-container-runtime-subordinates)

PR #473, opened 2019-08-22
Build Kata container charm

joedborg:
Adding Kata container charm to builds.


joedborg:
[Kata charm needs publishing to the Charm Store](https://trello.com/c/KmLvp8wD/1359-kata-charm-needs-publishing-to-the-charm-store)

PR #485, opened 2019-09-04
Changing agent label

hyperbolic2346:
Any node with the label runner can handle this job. Originally it was copy-pasted from another Jenkinsfile that controlled this with a parameter. No parameter on this side made it null and explode.


hyperbolic2346:
Merging this to get the test going

PR #499, opened 2019-09-09
Fix Vault test failing with kubectl not found

johnsca:
TBH, I'm not sure how this ever worked. `/snap/bin` isn't on the path by default (might depend on series of the deployed charm), and the config path has to be manually specified on the workers (but not on the masters).

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -1284,8 +1284,7 @@ async def test_encryption_at_rest(model, tools):
         await model.add_relation("kubernetes-master:vault-kv", "vault:secrets")
         await tools.juju_wait()
         # create secret
-        one_master = random.choice(model.applications["kubernetes-worker"].units)
-        await one_master.run("/usr/bin/snap install --edge client-keystone-auth")
+        one_master = random.choice(model.applications["kubernetes-master"].units)
         output = await one_master.run(
             "/snap/bin/kubectl create secret generic test-secret "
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -1284,14 +1284,19 @@ class AuditTimestampError(Exception):
         await model.add_relation("kubernetes-master:vault-kv", "vault:secrets")
         await tools.juju_wait()
         # create secret
-        worker = model.applications["kubernetes-worker"].units[0]
-        output = await worker.run(
-            "kubectl create secret generic test-secret "
+        one_master = random.choice(model.applications["kubernetes-worker"].units)
+        await one_master.run("/usr/bin/snap install --edge client-keystone-auth")
```
hyperbolic2346: do you want to install this here?

jobs/integration/validation.py
```diff
@@ -1284,14 +1284,19 @@ class AuditTimestampError(Exception):
         await model.add_relation("kubernetes-master:vault-kv", "vault:secrets")
         await tools.juju_wait()
         # create secret
-        worker = model.applications["kubernetes-worker"].units[0]
-        output = await worker.run(
-            "kubectl create secret generic test-secret "
+        one_master = random.choice(model.applications["kubernetes-worker"].units)
```
hyperbolic2346: one_worker = ?

PR #503, opened 2019-09-09
Fix secret value in encryption test not being b64 encoded

johnsca:
Secrets, when read back via kubectl, come out base64 encoded, so the test needs to do the same to validate it.

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -10,6 +10,7 @@ import pytest
 import juju
 from asyncio_extras import async_contextmanager
 from async_generator import yield_
+from base64 import b64encode
 from datetime import datetime
 from .logger import log
 from pprint import pformat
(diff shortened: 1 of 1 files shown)


johnsca:
Example:

```
ubuntu@cory-fu-bastion:~$ kubectl create secret generic test-secret --from-literal=username='secret-value'
secret/test-secret created
ubuntu@cory-fu-bastion:~$ kubectl get secret test-secret -o yaml
apiVersion: v1
data:
  username: c2VjcmV0LXZhbHVl
kind: Secret
metadata:
  creationTimestamp: "2019-09-09T22:05:51Z"
  name: test-secret
  namespace: default
  resourceVersion: "1348"
  selfLink: /api/v1/namespaces/default/secrets/test-secret
  uid: 4af92f65-54db-4460-a1fe-57541099eeae
type: Opaque
ubuntu@cory-fu-bastion:~$ echo c2VjcmV0LXZhbHVl | base64 -d
secret-valueubuntu@cory-fu-bastion:~$ echo -n 'secret-value' | base64
c2VjcmV0LXZhbHVl
```

PR #539, opened 2019-10-29
Disconnect latest stable from master branch

ktsakalozos:
With this PR we want to make sure that if there is a branch on GH for any track even the current stable (eg 1.16 at the time of this PR) that branch will be used in the LP builders.

The behavior we have until now is that the current stable and the latest stable LP builders build from the master GH branch.  

---

- For integration tests please make those changes in `jobs/integration`
- For MicroK8s, those changes are in `jobs/build-microk8s`


ktsakalozos:
Merging this to test it.

PR #542, opened 2019-10-29
Build microk8s with core18 support

knkski:


ktsakalozos:
Thank you @knkski lets try this

PR #555, opened 2019-12-02
Fix action results for juju 2.7.0

Cynerva:
Hopefully. I haven't tested this yet. If you want to wait a minute, I'll give it a quick run of validation.


Cynerva:
Validation passed locally.

PR #572, opened 2020-03-17
Add validate-multus job

Cynerva:
Still working on this, but open to early feedback.

So far I've added `test_multus` to validation.py, which depends on a new `addons_model` fixture and corresponding `--addons-model` option in conftest.py.

The goal is to have a validate-multus job that will deploy charmed-kubernetes with Ceph, add a k8s model, deploy Multus to the k8s model, and then run pytest with the new `--addons-model` option.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -37,6 +37,12 @@ def pytest_addoption(parser):
         required=False,
         help="Snap channel to use eg 1.16/edge",
     )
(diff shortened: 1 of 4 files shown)


battlemidget:
looks good so far, this is what i had in mind for usage of arguments to control that test flow

Cynerva:
This is ready for review. I've added the job definition for validate-ck-multus.

PR #576, opened 2020-03-26
Using `ctr` to do the pull, tag and push to ROCKS

joedborg:
Using `ctr` to do the pull, tag and push to ROCKS as `ctr` supports pulling all platforms at once

### DIFF
--- a/jobs/build-snaps/build-release-cdk-addons.groovy
+++ b/jobs/build-snaps/build-release-cdk-addons.groovy
@@ -123,8 +123,8 @@ pipeline {
         stage('Setup LXD container for ctr'){
             steps {
                 ssh "sudo lxc launch ubuntu:18.04 image-processor"
-                exec "image-processor" "apt update"
-                exec "image-processor" "apt install containerd -y"
+                lxd_exec "image-processor" "apt update"
+                lxd_exec "image-processor" "apt install containerd -y"
             }
         }
         stage('Process Images'){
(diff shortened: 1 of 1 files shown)


kwmonroe:
We're gonna hold this in draft until we're clear of the 1.18 release.  In addition to this piece, cdk-addons will need to *stop* injecting `multiarch_workaround` in the images, and we can't land that piece in `master` until 1.18 is safely out the door.

jobs/build-snaps/build-release-cdk-addons.groovy
```diff
@@ -158,31 +164,27 @@ pipeline {
...
-                            then
-                                # inject '-arch:' between the image name and version, as our templates expect
-                                RAW_IMAGE=\${RAW_IMAGE%%:*}-${params.arch}:\${RAW_IMAGE#*:}
-                                break
-                            fi
-                        done
 
                         # Tag and push
-                        docker tag \${i} \${TAG_PREFIX}/\${RAW_IMAGE}
+                        sudo lxc exec image-processor -- ctr image tag \${i} \${TAG_PREFIX}/\${RAW_IMAGE}
```
battlemidget, line 169: Whats the format here for adding the registry, I couldnt find anything in their source code or documentation

joedborg: We can, but I don't think we need to because the tag will contain the URL to the registry, so should just be a case of pushing with the creds (`--user"${USERNAME}:${PASSWORD}"`).  Or am I missing something?

battlemidget: Ah you're right I missed the tag_prefix had the URL in there.  With that said I think this was all we needed then?

joedborg: I think so too

PR #585, opened 2020-04-13
Add test for cdk-cinder storage class on OpenStack

johnsca:
Modified from the Ceph test, and marked to only run on OpenStack clouds.
The implementation for getting the cloud is incomplete.


johnsca:
Travis failures are lint errors unrelated to this change.

PR #602, opened 2020-06-09
Always set the MIcroK8s track when releasing

ktsakalozos:
Fixes the error:
```
jackal@aurora:~$ 'snapcraft' 'release' 'microk8s' '1443' 'stable'
invalid-field: Implicit track not allowed
```


ktsakalozos:
@battlemidget I missed that part  in may last PR so tests were passing and the release was failing with the above error.

I am merging this and queuing some jobs to make sure all is fine.

PR #604, opened 2020-06-09
Make get_last_audit_entry_date retry on JSONDecodeError

Cynerva:
Since we can't reproduce to investigate further, let's try this and hope it helps.


Cynerva:
Might fix https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1868130. I dunno.

PR #626, opened 2020-09-03
Log error when microk8s snap release fails

johnsca:
Debugging for CI failure


ktsakalozos:
LGTM +1 thank you @johnsca

PR #627, opened 2020-09-03
Change Vault test to use custom waiting

johnsca:
The unsealing of Vault blocks the cluster from settling properly, so it needs to be handled specially in the test.


kwmonroe (approved):
lgtm

PR #631, opened 2020-09-10
set kube_ersion early if we have a kube_version

kwmonroe:
If we specify a `kube_version` param, we'll never set the `kube_ersion`.  Fix that by setting it early when a `kube_version` is present.


kwmonroe:
This tested well [with a k8s_tag](https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-amd64-1.19/218/) and [without](https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-amd64-1.19/219/).  Both were dry runs; let's see how a [real one](https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-amd64-1.19/220/) goes.

kwmonroe (commented):
LGTM, thanks so much for the PR!

PR #636, opened 2020-09-18
Fix scp issues where wrong address is used

Cynerva:
libjuju's Unit.scp_from and Unit.scp_to don't properly handle some clouds (lxd, vsphere) where public addresses aren't discovered correctly.

This changes our scp code to use Juju CLI for all clouds, not just localhost.


Cynerva:
Fixes https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1888434

johnsca (commented):
LGTM

PR #637, opened 2020-09-23
Add metallb charms to support matrix

johnsca:
Uses override-build to use the new `make charm` pattern. This will be cleaned up in a future refactor. Also adds passes the name of the charm being built in, since the Makefile can build either charm from the same repo.

### DIFF
--- a/cilib/run.py
+++ b/cilib/run.py
@@ -22,11 +22,14 @@ def _log_sub_out(pipe):
 def script(script_data, **kwargs):
     is_single_command = len(script_data.splitlines()) == 1
     process = None
+    env = os.environ.copy()
+    if "charm" in kwargs:
+        env["CHARM"] = kwargs.pop("charm")
     if is_single_command:
         process = subprocess.Popen(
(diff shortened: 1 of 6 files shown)


cilib/run.py
```diff
@@ -41,7 +44,7 @@ def script(script_data, **kwargs):
     os.close(tmp_script[0])
     process = subprocess.Popen(
         ["bash", str(tmp_script_path)],
-        env=env.copy(),
+        env=os.env.copy(),
```
kwmonroe, line 47: Did you mean to use `env` or `env.copy()` or `os.environ.copy()` instead of `os.env.copy()`?

kwmonroe (changes requested):
I think you have a typo up in here.

cilib/run.py
johnsca: Yeah, that should just be `env`.  I don't think the previous version even worked, though.

kwmonroe (approved):

PR #641, opened 2020-10-01
Make Vault test use HA

johnsca:
Test Vault in an HA configuration without EasyRSA.

The TL;DR process for using Vault HA w/o EasyRSA is:

* Deploy Vault as per normal but with multiple units and a cert relation to etcd
* Wait for at least the Vault leader to settle
* Manually Init and unseal the leader unit
* Generate or provide a root CA cert to Vault
* Wait for the etcd units to settle
* Manually restart the `vault` service and (re)do the unseal on all of the Vault units

There are a couple of status issues in the k8s-master and vault charms which will be fixed with upcoming releases (including a potential hook error on k8s-master depending on the timing of when the vault services are restarted / re-unsealed due to https://github.com/juju-solutions/layer-vault-kv/pull/11) but these are just reporting issues which don't impact functionality (the errored hook, if encountered, can just be retried once the Vault units are back up).

Part of [lp:1897818](https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1897818)

### DIFF
--- a/jobs/integration/utils.py
+++ b/jobs/integration/utils.py
@@ -508,13 +508,16 @@ async def finish_series_upgrade(machine, tools):
 
 
 class JujuRunError(AssertionError):
-    def __init__(self, command, result):
+    def __init__(self, unit, command, result):
+        self.unit = unit
         self.command = command
         self.code = result.code
         self.stdout = result.stdout
         self.stderr = result.stderr
         self.output = result.output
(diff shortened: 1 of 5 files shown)


Cynerva (approved):

Cynerva (changes requested):
Looks like a merge conflict came up, can you resolve it?

Cynerva (approved):

PR #646, opened 2020-10-15
Refactor charm building to support charmcraft

johnsca:
With this, we can support charms in either style, driven off of the existence or not of a `layer.yaml` file, as well as handling both file and image type resources which must be custom built.

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -17,9 +17,7 @@ Usage:
 """
 
 import os
-from glob import glob
 from pathlib import Path
-from pprint import pformat
 from sh.contrib import git
 from cilib.service.aws import Store
 from cilib.run import cmd_ok, capture, script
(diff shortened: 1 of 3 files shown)


jobs/build-charms/charms.py
```diff
@@ -335,21 +371,26 @@ def setup(self):
...
             )
+        elif self.legacy_charm:
+            ret = cmd_ok(
+                "charm build -r --force -i https://localhost",
+                cwd=self.src_path,
+            )
         else:
             ret = cmd_ok(
-                "charm build -r --force -i https://localhost", cwd=self.src_path
+                f"{self.build.home_dir}/.local/bin/charmcraft build -f {self.src_path}",
```
hyperbolic2346, line 388: This hard coded path seems odd. Is there not a path to charmcraft available somewhere?

jobs/build-charms/resource-spec.yaml
```diff
@@ -1,27 +1,31 @@
 'cs:~containers/canal':
-  calico-amd64.tar.gz: calico
-  calico-arm64.tar.gz: calico-arm64
-  calico-upgrade-amd64.tar.gz: calico-upgrade
-  calico-upgrade-arm64.tar.gz: calico-upgrade-arm64
-  calico-node-image.tar.gz: calico-node-image
-  flannel-amd64.tar.gz: flannel
-  flannel-arm64.tar.gz: flannel-arm64
+  calico: '{out_path}/calico-amd64.tar.gz'
```
hyperbolic2346, line 2: this feels so much better.

jobs/build-charms/charms.py
adam-stokes: CI has se unique obstacles to deal with bc the home directory is not what snap can use. So we use pypi local installs which guarantee the path. I think this is fine

PR #649, opened 2020-10-23
Refactor DNS test to include CoreDNS charm

johnsca:
This also adds a k8s_model fixture for testing other K8s charms within the main job, although it looks like sriov and multus already have separate jobs as well.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -7,11 +7,15 @@ import asyncio
 import uuid
 import yaml
 import requests
+import click
 from datetime import datetime
+from pathlib import Path
 from py.xml import html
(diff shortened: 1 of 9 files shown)


jobs/integration/conftest.py
```diff
@@ -94,24 +98,40 @@ class Tools:
     """Utility class for accessing juju related tools"""
 
     def __init__(self, request):
+        self._request = request
         self.requests = aioify(obj=requests)
+
+    async def _load(self):
```
johnsca, line 104: I had to split this out to get the current Juju user, which was required for the fully-qualified model & offer names.

johnsca:
I didn't run the other tests with this, so I can do that on Monday before we merge it.

PR #656, opened 2020-11-06
use rocks images since not all ci infra can get upstream

kwmonroe:
Fixes https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/1903213

Some CI infra can't get out to docker.io/gcr.io/etc, but rocks.c.c is ok.  Replace upstream images with rocks equivalents.

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -1246,7 +1246,6 @@ async def test_audit_custom_policy(model, tools):
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip_arch(["aarch64", "s390x"])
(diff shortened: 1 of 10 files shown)


jobs/integration/validation.py
```diff
@@ -1246,7 +1246,6 @@ async def test_audit_custom_policy(model, tools):
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip_arch(["aarch64", "s390x"])
```
kwmonroe, line 1249: FYI, i added a multiarch nginx image to rocks, so we don't need to skip this test anymore.

PR #659, opened 2020-11-12
WIP: Do not release latest/stable from latest track

ktsakalozos:
Latest/stable should be releasing from <upstream_stable>/stable, eg 1.19/stable


ktsakalozos:
Gave it some testing, we should. Will bee keeping an eye on the upcoming releases.

PR #676, opened 2021-01-05
Add keepalived to CK release process

Cynerva:


jobs/includes/charm-support-matrix.inc
```diff
@@ -73,7 +73,7 @@
     upstream: "https://github.com/juju-solutions/charm-keepalived.git"
     downstream: 'juju-solutions/charm-keepalived.git'
```
Cynerva, line 74: I wasn't able to move the fork myself so the upstream/downstream are the same for now. If we do get it moved I'll open another PR to fix them.

Cynerva: Nevermind, the fork has been moved now so I've updated these in this PR.

PR #678, opened 2021-01-07
Skip building development versions if a non prerelease exists

adam-stokes:
During the next development phase, ie 1.21 we make sure to build all
prereleases (alpha, beta, rc). However, during the time we need to switch 1.21
to GA we do not want to build any additional prereleases after that. This is
because there is a short period of time where we block on 1.21 GA for testing
purposes before releasing to stable. During that wait period a 1.21.1-rc.0 may
be available and we do not want to build and have those end up in our stable
tracks during our automated release process.

This makes sure that if any non prerelease exists that we dont accidentally
build and push a non prerelease to stable. Once we bump the K8S_NEXT_VERSION to
1.22 it will pick back up and begin building those pre-releases for the next
development version.

Signed-off-by: Adam Stokes <51892+adam-stokes@users.noreply.github.com>

---

- For integration tests please make those changes in `jobs/integration`
- For MicroK8s, those changes are in `jobs/build-microk8s`

### DIFF
--- a/cilib/service/snap.py
+++ b/cilib/service/snap.py
@@ -134,6 +134,7 @@ class SnapService(DebugMixin):
                     track=f"{_version}/edge",
                     arch=arch,
                 )
+
                 if max_rev:
                     latest_snap_version = self.snap_model.store.version_from_rev(
                         max_rev, arch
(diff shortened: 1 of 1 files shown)


joedborg (approved):
Looks good to me

PR #679, opened 2021-01-08
Update validate-ck-calico and validate-ck-tigera-secure-ee to reuse VPCs

Cynerva:
This should fix the issues we're constantly seeing with VPC quota limits and cleanup failures. Instead of creating a new VPC each time a validate-ck-calico job runs, we'll have two long-lived VPCs that we reuse between jobs.

This PR removes `tigera_aws.py bootstrap` completely, and removes all calls to `tigera_aws.py cleanup` from the jobs. The VPC creation code has been moved over to a new `tigera_aws.py create-vpc` command that can be called manually by engineers as needed. I have already done this.

Included in this PR are several IPv6 fixes which I'll highlight inline.

### DIFF
--- a/jobs/integration/tigera_aws.py
+++ b/jobs/integration/tigera_aws.py
@@ -8,12 +8,14 @@ import tempfile
 import time
 import yaml
 import boto3
-from subprocess import check_output, check_call, CalledProcessError
+from subprocess import check_output, CalledProcessError
 
 VPC_CIDR = "172.30.0.0/16"
 SUBNET_CIDRS = ["172.30.0.0/24", "172.30.1.0/24"]
 REGION = "us-east-2"
 AVAILABILITY_ZONE = "us-east-2a"
+OWNER = "k8sci"
+
 BIRD_CONFIG_BASE = """
 log syslog all;
 debug protocols all;
@@ -85,10 +87,9 @@ def sh(*args, **kwargs):
 
 
 def tag_resource(resource_id):
-    owner = os.environ.get("JOB_NAME", "test-calico")
     ec2.create_tags(
         Resources=[resource_id],
-        Tags=[{"Key": "created-by", "Value": owner}],
+        Tags=[{"Key": "owner", "Value": OWNER}]
     )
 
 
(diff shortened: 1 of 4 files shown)


jobs/integration/tigera_aws.py
```diff
@@ -149,19 +145,28 @@ def bootstrap():
...
-    subnet_cidrs = SUBNET_CIDRS[:num_subnets]
-    for subnet_cidr in subnet_cidrs:
+    for i in range(num_subnets):
+        subnet_cidr = SUBNET_CIDRS[i]
+        subnet_ipv6_cidr = ipv6_subnet_cidrs[i]
         subnet_id = ec2.create_subnet(
             VpcId=vpc_id,
             CidrBlock=subnet_cidr,
-            Ipv6CidrBlock=ipv6_cidr_block,
+            Ipv6CidrBlock=subnet_ipv6_cidr,
```
Cynerva, line 169: This fixes an issue where the code tried to create multiple subnets with the same Ipv6CidrBlock, resulting in failure due to conflict. Now each subnet will get a unique IPv6 CIDR block.

jobs/integration/tigera_aws.py
```diff
@@ -188,40 +193,21 @@ def bootstrap():
             DestinationCidrBlock="0.0.0.0/0",
             GatewayId=gateway_id,
         )
+        ec2.create_route(
+            RouteTableId=route_table_id,
+            DestinationIpv6CidrBlock="::/0",
+            GatewayId=gateway_id,
+        )
```
Cynerva, line 200: This fixes public IPv6 servers (e.g. snap store) being unroutable.

jobs/integration/validation.py
```diff
@@ -574,9 +574,14 @@ async def get_to_restricted_networkpolicy_service():
...
 
 @pytest.mark.asyncio
-@pytest.mark.skip_apps(["calico"])
 async def test_ipv6(model, tools):
-    master = model.applications["kubernetes-master"].units[0]
+    master_app = model.applications["kubernetes-master"]
+    master_config = await master_app.get_config()
+    service_cidr = master_config["service-cidr"]["value"]
+    if all(ipaddress.ip_network(cidr).version != 6 for cidr in service_cidr.split(',')):
+        pytest.skip("kubernetes-master not configured for IPv6")
```
Cynerva, line 582: Calico doesn't support IPv6 + VXLAN, so we have to disable IPv6 when testing VXLAN. I had to make the test_ipv6 skip logic a bit smarter to accomodate that.

PR #690, opened 2021-02-11
Use oci image staging area during the cdk-addons build job

kwmonroe:
To prevent dockerhub rate limits, we're splitting the upstream image pull logic into a new job that runs less frequently than our cdk-addons build job.  That job will pull upstream images and tag/push them to a rocks/staging area.

This PR changes the origin for all upstream images to rocks/staging, which means we can drop our docker login bits. Our staging area should *always* contain the images that cdk-addons needs, so i removed the error handling around pull/tag. If we have a problem there, we want this build job to fail until we figure out what's missing in staging.

Finally, I bumped up the image-processor lxc container to use a focal image (why not?) and updated the `snapcraft` commands to use `upload` instead of the deprecated `push`.


kwmonroe (commented):
Jobs are running successfully from this branch, so i'm gonna go ahead and merge to master.

PR #693, opened 2021-03-08
Fix pre-release microk8s tags

ktsakalozos:
It seems the "name" field in https://api.github.com/repos/kubernetes/kubernetes/releases changed to include a "Kubernetes" at the beginning. We switch "to tag_name" with this PR to get the release tag name.


joedborg (approved):
Looks good to me

PR #700, opened 2021-03-30
update sync / cdk-addons jobs

kwmonroe:
Syncing images takes a lot (25G) of disk.  Add debug to monitor this disk usage, and set the appropriate jobs to use our `bigdisk` machines.

Drive-by update to the sync-oci job to set the default k8s version to 1.21.

### DIFF
--- a/jobs/build-snaps.yaml
+++ b/jobs/build-snaps.yaml
@@ -21,7 +21,7 @@
           default: '{arch}'
       - string:
           name: build_node
-          default: 'runner-{arch}'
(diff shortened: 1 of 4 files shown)


Cynerva (approved):

johnsca (approved):
LGTM.  If the `df` output gets noisy from snaps / loop devices, something like this might be useful:

```
df -h -x squashfs -x overlay | grep -vE " /snap|^tmpfs|^shm"
```

PR #703, opened 2021-04-01
report what images are used in a ck release

kwmonroe:
We have a [list of container images](https://github.com/charmed-kubernetes/bundle/blob/master/container-images.txt) that gets processed/updated during `build-cdk-addons`.  It serves our build jobs well, but it isn't very human-readable -- `static` and `upstream` lines need to be combined, image names come directly from upstream templates, and items like `{{ arch }}` aren't expanded.

During each build, we end up massaging images into a format that would be way easier for humans to grok. This PR adds these to a per-release text file that gets committed [to the same repo](https://github.com/charmed-kubernetes/bundle/tree/master/container-images).

Drive-by to quiet the extreme log noise when pulling/pushing images.

Fixes https://bugs.launchpad.net/cdk-addons/+bug/1891530 by making it easier to call out image changes in release notes.

### DIFF
--- a/jobs/build-snaps/build-release-cdk-addons.groovy
+++ b/jobs/build-snaps/build-release-cdk-addons.groovy
@@ -236,8 +236,7 @@ pipeline {
                     # Commit what we know about our images
                     cd bundle
                     REPORT_FILE=container-images/${kube_version}.txt
-                    REPORT_IMAGES=\$(echo "\${REPORT_IMAGES}" | xargs -n1 | sort -u)
-                    echo \${REPORT_IMAGES} > \${REPORT_FILE}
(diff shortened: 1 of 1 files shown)


kwmonroe:
Successful run with this PR:

https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-amd64-1.19/391/

And resultant output committed to bundle repo:

https://github.com/charmed-kubernetes/bundle/blob/master/container-images/v1.19.9.txt

kwmonroe:
This has been working well for a few days with the overnight `build-cdk-addons` jobs on this branch.  Merging to master.

Cynerva (commented):
LGTM

PR #704, opened 2021-04-09
CI support for 1.21

kwmonroe:
Drop 1.17 from our test matrix; add 1.21.

For new deploy tests, drop xenial; new deployments should be >= bionic. We still support ESM customers upgrading from < 1.21/xenial to 1.21/xenial, so we'll keep targeting the upgrade tests for >= xenial.

Move unused jobs to our `.deprecated` dir, just for convenience in case they need to be revived.


kwmonroe:
Jobs are running as expected, but we now need to do a 1.22 swizzle to get the 1.21 GA bits in place.  Merging this to master so we can keep 1.21-related changes separate from 1.22-alpha.

PR #707, opened 2021-04-16
tweak for 1.21 being the new stable

kwmonroe:
Set `K8S_STABLE` to 1.21.  With this, `sync-snaps` will promote `1.21/stable` snaps to `latest/stable`. 

Set proper `go` and ppa config for 1.22.


kwmonroe:
I went ahead and swizzled the snap versions for conformance, release, and validation tests in this PR rather than making individual PRs for each of those.

1.21 is now our default, meaning anything <= 1.18 is out of support.  I'm also dropping `xenial` tests for base-deployment tests (we still keep it for upgrade tests).

jobs/release.yaml
```diff
@@ -54,9 +54,9 @@
           type: user-defined
           name: snap_version
           values:
-            - 1.20/edge
-            - 1.19/edge
-            - 1.18/edge
+            - 1.21/stable
+            - 1.20/stable
+            - 1.19/stable
```
Cynerva, line 59: I moved these to edge [here](https://github.com/charmed-kubernetes/jenkins/pull/691) because I noticed during the last bugfix release that we were about to release cdk-addons changes completely untested. We can't revert this back to stable staps unless we have cdk-addons test coverage somewhere else in the bugfix release process.

Cynerva (changes requested):

jobs/release.yaml
kwmonroe: After discussing our need to test with a pre-stable `cdk-addons`, and based on the fact that `cdk-addons` publishes edge through candidate on each build. i've moved these to `1.xx/candidate`.  Hopefully it's more clear that `validate-charm-bugfix` is going to use candidate charms with candidate snaps now.

Cynerva (approved):

PR #708, opened 2021-04-22
Focal updates

kwmonroe:
Jenkins workers were updated to focal.  We need to adjust packages and runner config to work with them.

### DIFF
--- a/jobs/build-snaps.yaml
+++ b/jobs/build-snaps.yaml
@@ -21,7 +21,7 @@
           default: '{arch}'
       - string:
           name: build_node
-          default: 'runner-amd64-bigdisk'
+          default: 'runner-amd64'
       - string:
           name: version
           default: '{version}'
--- a/jobs/sync-oci-images.yaml
+++ b/jobs/sync-oci-images.yaml
@@ -13,10 +13,10 @@
     parameters:
       - string:
           name: build_node
-          default: 'runner-amd64-bigdisk'
+          default: 'runner-amd64'
       - string:
           name: version
-          default: '1.21'
+          default: '1.22'
           description: |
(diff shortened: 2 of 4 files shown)


kwmonroe:
The most critical piece of this PR is to our infra jobs; this tested successfully with https://jenkins.canonical.com/k8s/job/infra-maintain-nodes-amd64-0/.

There may be something like a py36 lurking in some other job that needs an update, but they won't be easy to find until we run this new `infra-maintain-nodes` across all workers (because atm, *all* jobs will fail on these workers).

jobs/build-snaps.yaml
```diff
@@ -21,7 +21,7 @@
           default: '{arch}'
       - string:
           name: build_node
-          default: 'runner-amd64-bigdisk'
+          default: 'runner-amd64'
```
Cynerva, line 24: bigdisk no longer needed? How come?

jobs/sync-oci-images.yaml
```diff
@@ -13,10 +13,10 @@
     parameters:
       - string:
           name: build_node
-          default: 'runner-amd64-bigdisk'
+          default: 'runner-amd64'
```
Cynerva, line 16: bigdisk no longer needed? How come?

Cynerva (approved):
I don't understand why the bigdisk changes but I don't wanna hold up this PR over that. Looks good.

jobs/build-snaps.yaml
kwmonroe: We actually got new workers instead of "upgraded" ones.  They all have 50+ gb root disks, so there's no need to make this distinction -- sync and snap jobs can run on any amd64 workers.

PR #709, opened 2021-04-26
Docker fixes

kwmonroe:
Update the infra jobs to append required groups to the existing `jenkins` user.

Infra jobs call `docker prune` and run periodically throughout the day. We previously forced pruning on `build-snaps` and `sync-x` jobs to save disk space, but it's not required with our current jenkins workers since they have larger disks.

### DIFF
--- a/jobs/infra/playbook-jenkins.yml
+++ b/jobs/infra/playbook-jenkins.yml
@@ -342,7 +342,8 @@
     - name: update jenkins user
       user:
         name: jenkins
-        groups: jenkins,docker,lxd
+        groups: docker,lxd
+        append: yes
       tags:
         - jenkins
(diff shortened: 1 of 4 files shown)


jobs/infra/playbook-jenkins.yml
```diff
@@ -342,7 +342,8 @@
     - name: update jenkins user
       user:
         name: jenkins
-        groups: jenkins,docker,lxd
+        groups: docker,lxd
+        append: yes
```
kwmonroe, line 346: you may notice we drop `jenkins` from groups, but that's our primary group, so we don't need to add it in the `groups` list; we just need to make sure we append the others.

Cynerva (approved):
LGTM

PR #714, opened 2021-06-28
Lessen sync-oci-images disk requirements

kwmonroe:
The sync-oci-images job pulls, tags, and pushes around 100 images per job.  This takes around 30G, which can fill up our worker root disks, especially if other jobs are running concurrently. There's no good reason to keep processed images around for the duration of the job -- we destroy the container used to fetch them at the end of the job anyway.

This PR removes an image as soon as we're done pushing it to rocks.  It also sends `ctr push/pull` output to `/dev/null` because the level of output is enormous, and even if a `pull/push` fails mid-way through an image, that info isn't useful.  We care about *what* image failed, not *when* it failed.

### DIFF
--- a/jobs/sync-oci-images/sync-oci-images.groovy
+++ b/jobs/sync-oci-images/sync-oci-images.groovy
@@ -69,8 +69,8 @@ pipeline {
                         git clone https://github.com/charmed-kubernetes/cdk-addons.git --depth 1
                     fi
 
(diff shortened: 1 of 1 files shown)


kwmonroe:
Example job based on this branch: https://jenkins.canonical.com/k8s/job/sync-oci-images/152/console

Cynerva (approved):

PR #715, opened 2021-06-28
use build_number in the lxc name for uniqueness

kwmonroe:
If multiple `build-release-cdk-addons-x` jobs run on the same host, the first one to finish will delete the `image-processor` lxd container. This will cause all other jobs running on that worker to fail.

Add the build number to the container name to make it unique.


kwmonroe:
Successful run with this branch: https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-amd64-1.20/256/console

Cynerva (approved):

PR #717, opened 2021-07-12
set model defaults to auto retry hooks

kwmonroe:
Our [CI reports](http://jenkaas.s3-website-us-east-1.amazonaws.com/) are littered with yellow nightly jobs; some of these can be attributed to charmstore/snapstore failures (e.g 503 errors when trying to pull a snap). We currently fail a test with "yellow" status even if things outside our control (snap/charmstore) are down because we have "auto-retry-hooks" turned off in our CI models.

This PR proposes we enable retry-hooks to get around transient network issues with the assumption that SQA is our release gate -- they test with `auto-retry-hooks=false`. This will enable us to easily see if our tests are failing *with* a hook-retry (via our nightly job status of red/yellow/green), yet still let SQA hold a release for a failure that might be a legit issue only surfaced when `auto-retry-hooks=false`.

Without this, our CI report is so yellow that we've lost confidence in it. I think our daily CI should tolerate external failures. The risk here is that we deliberately turned this option off to better match SQA's env, but in hindsight, I think this has led to us ignoring CI failures with the mindset of "eh, probably a network error; let's see how tomorrow goes".  But then tomorrow comes and we fail differently and rinse/repeat.

### DIFF
--- a/jobs/bundles/test-all
+++ b/jobs/bundles/test-all
@@ -77,7 +77,7 @@ function juju::bootstrap
          --model-default test-mode=true \
          --model-default resource-tags=owner=k8sci \
          --model-default image-stream=daily \
-         --model-default automatically-retry-hooks=false \
+         --model-default automatically-retry-hooks=true \
          --model-default logging-config="<root>=DEBUG"
 }
 
--- a/jobs/validate/calico-spec
+++ b/jobs/validate/calico-spec
@@ -33,7 +33,7 @@ function juju::bootstrap
       --model-default test-mode=true \
       --model-default resource-tags=owner=k8sci \
       --model-default image-stream=daily \
-      --model-default automatically-retry-hooks=false \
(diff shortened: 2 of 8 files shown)


johnsca:
Hook errors, even if they resolve themselves, seem like a bad user experience.  If we're running up against transient network issues, we should handle those and retry some reasonable number of times and then set a blocked status with an appropriate message and then retry again on a later hook.  If we're still getting failures due to these network or outage issues, then we need to figure out how to improve the reliability of the network or services or escalate it to those who can.

johnsca (approved):
Per further discussion, I'm +1 on this given the caveat that we add additional logic to the log parsing & reporting code to detect when retries led to success so that those cases don't get missed.  Doing so will let us both detect failures further down the line as well as tell when a transient error needs to be addressed.

PR #718, opened 2021-07-27
Fix timed out tests not showing tracebacks, use SIGINT

Cynerva:
If we send SIGINT when a timeout occurs, that will raise a KeyboardInterrupt exception in the Python code and show us a traceback. That should help us see exactly where tests are getting stuck.

### DIFF
--- a/ci.bash
+++ b/ci.bash
@@ -128,6 +128,7 @@ function test::execute
     declare -n is_pass=$1
     timeout -s INT 3h pytest \
         --html="report.html" \
+        --full-trace \
         jobs/integration/validation.py \
         --cloud "$JUJU_CLOUD" \
         --model "$JUJU_MODEL" \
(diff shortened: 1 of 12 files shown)


Cynerva:
I did `git grep 'timeout.*pytest'` and just added `-s INT` to every instance I saw.

johnsca (changes requested):
I tested the `timeout -s INT` and it seems that pytest will hide the traceback specifically for `KeyboardInterrupt`:

```
/home/johnsca/juju/envs/dev/charms/kubernetes-master/tests/unit/test_kubernetes_master.py:66: KeyboardInterrupt
(to show a full traceback on KeyboardInterrupt use --full-trace)
```

I think we need to add the `--full-trace`.  Adding that gave the trace info in my test, but it's worth noting that `--full-trace` seems to override `--tb=native` but I think actually having the debug info is worth the more verbose traceback style.

Cynerva:
Good catch, thanks. I've added the `--full-trace` arg to pytest now too.

johnsca (approved):
LGTM

PR #719, opened 2021-08-04
Service endpoint unit print

asbalderson:
During test_clusterip_service_endpoint log k8s unit failures.
    
Test_clusterip_service_endpoint tries to reach the hello world
container from each of the k8s master and workers, this patch
will print which unit is running the curl to help debug any
errors.


johnsca (approved):

johnsca:
Thanks for this!  Apologies for the delayed review.

PR #721, opened 2021-08-13
fix channel map issues

kwmonroe:
surl-creds had expired.  Add a note for how to update these and print debug info if the channel-map appears invalid.  These credentials need to be force-written when infra jobs run, otherwise the playbook will not overwrite the creds file if one already exists.

Drive by to disable the sync-ppas job, since we no longer support k8s deb packages.

Edit: also fixed some lint errors in snapapi.py

### DIFF
--- a/jobs/sync-upstream.yaml
+++ b/jobs/sync-upstream.yaml
@@ -24,7 +24,7 @@
 
       Then upload the launchpad_creds file to the launchpad_creds credential in Jenkins.
 
-      If the job fails to retrive snap channel information, you may need to
+      If the job fails to retrieve snap channel information (e.g., you get an "invalid channel map" error), you may need to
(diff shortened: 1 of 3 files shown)


jobs/sync-upstream.yaml
```diff
@@ -22,7 +22,17 @@
       ...
       >>> exit()
 
-      Then upload the launchpad_creds file to the launchpad_creds credential on Jenkins.
+      Then upload the launchpad_creds file to the launchpad_creds credential in Jenkins.
+
+      If the job fails to retrive snap channel information, you may need to
```
johnsca, line 27:
```suggestion
      If the job fails to retrieve snap channel information (e.g., you get an "invalid channel map" error), you may need to
```

PR #723, opened 2021-08-16
add 1.22 to test matrix; adjust enums for proper snap builds

kwmonroe:
- Start pushing 1.22 snaps to `beta` and `candidate` in preparation for GA and bugfix releases.
- Drop 1.18 from our test matrix (it is out of support)
- Add 1.22 to our test matrix
- For upgrade jobs, make sure we include 1.21/stable in the "upgrade-from" matrix


johnsca (approved):
Looks reasonable to me except for the fact that so many places needing to be manually updated means I have no way of knowing if any of them got missed.

PR #724, opened 2021-08-17
CK-124 - Add additional debugging info to test_sans

johnsca:
The test_sans case doesn't provide much info to debug what has gone wrong. This adds some additional logging to assist with future failures.


kwmonroe (approved):
lgtm, thanks!

PR #725, opened 2021-08-18
update snap-eks jobs with arch/dry-run params

kwmonroe:
CPC has requested we build snaps for both amd64 and arm64.  Update the job with an arch param so we can set this per-run.

While we're here, update default version values to the latest that cpc has requested.

### DIFF
--- a/jobs/build-snaps.yaml
+++ b/jobs/build-snaps.yaml
@@ -92,15 +92,15 @@
       - string:
           name: ARCH
           default: 'amd64'
-          description: snap eks architecture to build/promote
+          description: eks snap architecture to build/promote
       - string:
           name: VERSION
           default: '1.21.1'
-          description: snap eks version to build/promote
+          description: eks snap version to build/promote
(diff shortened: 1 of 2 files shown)


kwmonroe:
I've updated the CI job to use this branch and it looks good per https://jenkins.canonical.com/k8s/job/build-release-eks-snaps/66/console:

```
...
15:54:23 Running: snapcraft upload release/snap/build/kubectl-eks_1.21.2_amd64.snap
15:54:25 Preparing to upload 'kubectl-eks_1.21.2_amd64.snap'.
15:54:25 After uploading, the resulting snap revision will be released to '1.21.2/beta', '1.21.2/candidate', '1.21.2/edge', and '1.21.2/stable' when it passes the Snap Store review.
15:54:25 Generating delta for 'kubectl-eks_1.21.2_amd64.snap'.
15:54:44 released
...
```

johnsca (approved):

PR #726, opened 2021-08-20
fix up test_ceph

kwmonroe:
`test_ceph` was failing on 1.22 because we needed cephcsi v3.3.  That landed with:

https://github.com/charmed-kubernetes/cdk-addons/pull/204

The test itself also has issues -- specifically when trying to verify that pods were deleted.  We checked for `"error" in output.results.get("Stdout", ""):`, and nowadays, that always matches because we have this tidbit in stdout:

```
...
"--handle-volume-inuse-error=false"
...
```

Refactor `verify_deleted` to be more like `verify_[completed|ready]` and utilize the `find_entitites` function.

### DIFF
--- a/jobs/integration/utils.py
+++ b/jobs/integration/utils.py
@@ -217,27 +217,16 @@ async def disable_source_dest_check(model_name):
     await asyncify(check_call)(cmd, env=env)
 
 
-async def verify_deleted(unit, entity_type, name, extra_args=""):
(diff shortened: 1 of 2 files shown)


kwmonroe:
Tested `validate-ck-amd64` with this branch.  Looks good:

http://jenkaas.s3-website-us-east-1.amazonaws.com/

PR #731, opened 2021-09-30
Fix Cinder test and add Octavia test for OpenStack

johnsca:
The filter for running tests on only specific clouds was driven by the `--cloud` param rather than the actual cloud the test is being run on, leading to tests being skipped if the cloud param wasn't given (defaulting to "aws").
    
This also adds a test for Octavia when running on OpenStack.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -16,6 +16,7 @@ from juju.model import Model
 from aioify import aioify
 from traceback import format_exc
 from .utils import upgrade_charms, upgrade_snaps, arch, log_snap_versions, scp_from
+from .logger import log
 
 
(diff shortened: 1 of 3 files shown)


johnsca:
I got a passing run for the Cinder test on 1.21, and on 1.22 with the updated cloud-provider-openstack.

I haven't been able to get a passing run for the Octavia test yet but AFAICT that's just an environment / config issue.  The LB gets created and assigned an IP, but it's not reachable from the instance running the test. That should mostly come down to me passing the right values in for the `subnet-id` and `floating-network-id` configs or setting up routing rules on the host instance, so that doesn't seem like it should be a blocker for this.

Cynerva (approved):

PR #734, opened 2021-10-19
Remove .deprecated folder

Cynerva:
We don't really need this hanging around, do we? It's adding noise to my `git grep` output.


kwmonroe (commented):
I'm a little nervous dropping these.  Our last couple releases have seen lots of CI trouble. I fear that some of these deprecated jobs will hold a clue about how jobs *used* to work to help debug future failures.

I'd like us to get another release or two under our belts to have confidence that we won't need any of these deprecated bits.

PR #735, opened 2021-10-20
Add bugfix release doc step for building cdk-addons

Cynerva:


kwmonroe (approved):
LGTM, but these build jobs only push through edge/beta/candidate by default.  That's fine for validating a bugfix release (which pulls from the candidate channel).

Later in this doc, we say `Promote cdk-addons snap from edge to proper snap channels for ck release.`  Since you're in here, consider updating that wording to say `...from candidate to the stable channel...`, which you would either do by re-running `build-release-cdk-addons-xxx` with `1.xx/stable` in the channel list (not sure if this would generate a new snap rev if no other code changed) or by promoting candidate snaps to stable in the snapcraft.io UI.

Again, the current doc changes are +1 from me, but it may be worth expanding on how to get those into `1.xx/stable` post-validation.

PR #741, opened 2021-11-08
Better image reporting

kwmonroe:
We skip processing images that we already host in rocks.  Unfortunately, this meant they weren't showing up in our images.txt files.  For example, note addon-resizer isn't in this list:

https://github.com/charmed-kubernetes/bundle/blob/master/container-images/v1.19.9.txt

This PR generates reported images *before* processing.  It also adds the full image path, e.g.:
```
rocks.canonical.com/cdk/cdkbot/microbot-amd64:latest
```
vs
```
cdkbot/microbot-amd64:latest
```

I opted to include the full path because i think it will be easier for folks to copy/paste usable URLs rather than having to prepend rocks/cdk to everything.

### DIFF
--- a/jobs/build-snaps/build-release-cdk-addons.groovy
+++ b/jobs/build-snaps/build-release-cdk-addons.groovy
@@ -184,12 +184,6 @@ pipeline {
 
                     for i in \${ALL_IMAGES}
                     do
-                        # Skip images that we already host
-                        if echo \${i} | grep -qi -e 'rocks.canonical.com' -e 'image-registry.canonical.com'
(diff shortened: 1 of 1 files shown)


kwmonroe:
Before:
https://github.com/charmed-kubernetes/bundle/blob/master/container-images/v1.23.0-alpha.3.txt

After:
https://github.com/charmed-kubernetes/bundle/blob/master/container-images/v1.23.0-alpha.4.txt

evilnick (approved):
:cake:

PR #739, opened 2021-11-05
Add build-microk8s-charm job definition

neoaggelos:
### Summary

Add Jenkins job for building MicroK8s charm and uploading to CharmHub

PR #742, opened 2021-11-08
MicroK8s charm Jenkins Jobs

neoaggelos:
### Summary

Add Jenkins jobs for running integration tests for the MicroK8s charm and releasing to CharmHub.

Also builds on top of the job introduced in #739, so that all builds are automatically released to the edge channel. 

### Notes

This PR implies https://github.com/canonical/charm-microk8s/pull/12

### DIFF
--- a/jobs/build-microk8s-charm.yaml
+++ b/jobs/build-microk8s-charm.yaml
@@ -54,3 +54,84 @@
 
           git clone https://github.com/charmed-kubernetes/jenkins.git -b master jenkins
           bash -eux jenkins/jobs/microk8s/charms/build.sh
+
+- job:
+    name: 'release-microk8s-charm'
+    description: |
+      Run integration tests on the MicroK8s charm and release to stable.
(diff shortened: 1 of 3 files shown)


neoaggelos:
cc @ktsakalozos

jobs/build-microk8s-charm.yaml
```diff
@@ -49,15 +54,31 @@
           [[ -f $WORKSPACE/.env ]] && source $WORKSPACE/.env
           set +o allexport
 
+          container="${JOB_NAME}-${BUILD_NUMBER}"
```
ktsakalozos, line 57: If the job does not finish cleanly or gets stuck and we have to kill it, would we be leaking containers in this way?

neoaggelos: Added an initial step that cleans up any left-over containers from previous runs

ktsakalozos (approved):
LGTM +1

PR #743, opened 2021-11-08
handle building in py3.8 and py3.9 environments

addyess:
Apply requirements.txt constraints depending on whether your in a < py39 or >=py39 environment

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -312,7 +312,7 @@ pymacaroons==0.13.0
     #   surl
 pymdown-extensions==8.1.1
     # via mkdocs-material
-pynacl==1.1.2
+pynacl==1.2.0
     # via
     #   -r requirements.in
     #   macaroonbakery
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -312,7 +312,7 @@ pymacaroons==0.13.0
     #   surl
 pymdown-extensions==8.1.1
     # via mkdocs-material
-pynacl==1.1.2
+pynacl==1.2.0
```
addyess, line 315: Bumped due to this issue https://github.com/freedomofpress/securedrop/issues/6023

requirements.txt
```diff
@@ -483,7 +483,8 @@ wcwidth==0.2.5
     # via
     #   prettytable
     #   prompt-toolkit
-websockets==7.0
+websockets==7.0 ; python_version<"3.9"
+websockets==8.0 ; python_version>="3.9"
```
addyess, line 487: Similar fix as in [libjuju](https://github.com/juju/python-libjuju/blob/master/setup.py#L38-L39)

addyess (commented):
I'm not totally sure this is the right way to make these changed, but they looked reasonable

kwmonroe (approved):

PR #744, opened 2021-11-15
Unittest: jobs/build-charms/charms to operate against charmstore and charmhub

addyess:
- New Unit tests of `jobs/build-charms/charms.py`
- updated `BuildEnn:promote_all` to allow for charmhub & charmstore promotion
- updated `BuildEntity:setup` to operate with charmhub and & charmstore
- updated `BuildEntity:charm_build` method
- updated `BuildEntity:has_changed` method
- updated `BuildEntity:promote` with charmhub and & charmstore
- updated `BuildEntity:push` with charmhub and & charmstore
- updated `BuildEntity:attach_resources` with charmhub and & charmstore
- updated `BundleBuildEntity:push` with charmhub and & charmstore
- updated `BundleBuildEntity:bundle_build`
- Unit testing of build command (charm and bundles).
- Unit testing of promote command.
- Real world testing locally to push builds to charmhub against `kubernetes-control-plane`
- Real world testing locally to push builds to charmstore against `kubernetes-master`


jobs/build-charms/charms.py
```diff
@@ -53,20 +53,23 @@ class LayerType(Enum):
 
 class BuildEnv:
     """Charm or Bundle build data class"""
+    CHARMCRAFT_RESOURCE = re.compile(r"(\S+) \(r(\d+)\)")
 
-    try:
```
addyess, line 57: Moved this to `__new__` method so it wasn't run on module load

jobs/build-charms/charms.py
```diff
@@ -157,40 +160,83 @@ def save(self):
...
+                        "--resource",
+                        "{}-{}".format(resource["name"], resource["revision"]),
+                    )
+                    for resource in resources
+                ]
+        except sh.ErrorReturnCode_1:
+            click.echo("No resources for {}".format(charm_id))
+        sh.charm.release(
+            charm_id["id"]["Id"], "--channel", to_channel, *resources_args
+        )
```
addyess, line 191: handles promotion of charms on the charmstore

jobs/build-charms/charms.py
```diff
@@ -157,40 +160,83 @@ def save(self):
...
+                row[key] = value or channel_status[-1].get(key)
+            resources = row.get('Resources', '')
+            if resources == '-':
+                row['Resources'] = []
+            elif resources == '↑':
+                row['Resources'] = channel_status[-1]['Resources']
+            else:
+                row['Resources'] = self.CHARMCRAFT_RESOURCE.findall(resources)
+            channel_status.append(row)
+        return channel_status
```
addyess, line 212: helper to parse output of `charmcraft status <charm>` because it doesn't provide a yaml format output

addyess:
The CI failure here is EXACTLY https://stackoverflow.com/questions/69912264/python-3-9-8-fails-using-black-and-importing-typed-ast-ast3

i could move forward in black to resolve -- but feeling iffy about that.

addyess:
Build failures related to https://github.com/python/typed_ast/issues/169

kwmonroe (commented):
I've been reviewing this as it has progressed, and it looks really good.  That said, I'd like to wait another day before merging so we can get the 1.23/beta charms in place (just in case `build-charms` changes don't quite work).

That will let us turn over charms/bundles to SQA for 1.23 testing and buy us some time in case errors crop up in the edge channels.

jobs/build-charms/charms.py
```diff
@@ -39,46 +41,299 @@
...
+                format="yaml",
+                _out=self._echo,
+            )
+            return yaml.safe_load(resources.stdout.decode())
+        except sh.ErrorReturnCode:
+            return []
+
+    def promote(self, charm_entity, from_channel, to_channel):
+        self._echo(
+            f"Promoting :: {charm_entity:^35} :: from:{from_channel} to: {to_channel}"
```
stonepreston, line 95:
```suggestion
            f"Promoting :: {charm_entity:^35} :: from: {from_channel} to: {to_channel}"
```

jobs/build-charms/charms.py
```diff
@@ -39,46 +41,299 @@
...
+            "actions": [
+                {
+                    "name": name,
+                    "base": {
+                        "name": "ubuntu",
+                        "architecture": architecture,
+                        "channel": base_channel,
+                    },
+                    "channel": channel,
+                    "action": "install",
```
stonepreston, line 188:
```suggestion
                    "action": "refresh",
```
Should this be refresh here?

addyess: i dunno?  do you have some evidence?  I can be convinced.

stonepreston: After digging into the [docs](https://api.snapcraft.io/docs/refresh.html) a bit they seem to suggest using the "download" action, which is the same as the install action, but is reserved for when you don't actually install anything. It does not really say why this distinction needs to be made however ¯\_(ツ)_/¯

jobs/build-charms/charms.py
```diff
@@ -39,46 +42,306 @@
...
+                }
+            ],
+        }
+        resp = requests.post("https://api.charmhub.io/v2/charms/refresh", json=data)
+        return resp.json()
+
+    def status(self, charm_entity):
+        """Read CLI Table output from charmcraft status and parse."""
+        charm_status = sh.charmcraft.status(charm_entity, _out=self._echo)
+        header, *body = charm_status.stdout.decode().splitlines()
```
stonepreston, line 195: I am trying some of the parsing locally and the list that is the result of splitlines is empty if _out=self._echo is included in the status sh call. If the call is simply `sh.charmcraft.status(charm_entity) then the list is returned as expected.

addyess: what if you remove `_out` and replace with `_tee=True`?

```suggestion
    def status(self, charm_entity):
        """Read CLI Table output from charmcraft status and parse."""
        charm_status = sh.charmcraft.status(charm_entity, _tee=True)
        header, *body = charm_status.stdout.decode().splitlines()
```

stonepreston: > what if you remove `_out` and replace with `_tee=True`?

The list looks correct with that

stonepreston: but again, this is using charmcraft 1.4.0, which might be the issue

addyess: what if you have both `_tee` and `_out`?

stonepreston: now we're talkin. List looks good and the output is shown.

jobs/build-charms/charms.py
```diff
@@ -681,19 +696,11 @@ def charm_build(self):
...
+            ret = _CharmStore(self).build(*args.split(), _cwd=self.src_path)
         else:
             args = f"-f {self.src_path}"
             self.echo(f"Building with: charmcraft build {args}")
-            try:
-                _CharmHub(self).build(*args.split(), _cwd=self.build.build_dir)
-            except sh.ErrorReturnCode:
-                ret.ok = False
-
+            ret = _CharmHub(self).build(*args.split(), _cwd=self.build.build_dir)
```
stonepreston, line 703: is _CharmStore used to build legacy charms for both the charm store and the hub?

jobs/build-charms/charms.py
```diff
@@ -681,19 +696,11 @@ def charm_build(self):
                 args += " --charm-file"
                 self.dst_path = str(Path(self.src_path) / f"{self.name}.charm")
             self.echo(f"Building with: charm build {args}")
-            try:
-                _CharmStore(self).build(*args.split(), _cwd=self.src_path)
-            except sh.ErrorReturnCode:
-                ret.ok = False
-
+            ret = _CharmStore(self).build(*args.split(), _cwd=self.src_path)
```
stonepreston, line 699: is _CharmStore used to build legacy charms for both the charm store and the hub?

addyess: this is obviously unclear.  
The command `charm` is used to build legacy charms (charms with a layer.yaml) regardless of which "store" they end up in.  if the target store is "charmhub" then you have to add `--charm-file`.  I wonder if i can rework this.  I'll test if `charmcraft` can build legacy charms now.  

The command `charmcraft` is used to build operator charms (newest version)

addyess: yeah, `charmcraft` wasn't able to build a reactive charm.  But i COULD update this function to look like this:

What do you think @stonepreston 

```python
    def charm_build(self):
        """Perform a build against charm/bundle."""
        if "override-build" in self.opts:
            self.echo("Override build found, running in place of charm build.")
            ret = script(
                self.opts["override-build"],
                cwd=self.src_path,
                charm=self.name,
                echo=self.echo,
            )
        elif self.legacy_charm:
            args = "-r --force -i https://localhost"
            if self.store == "ch":
                args += " --charm-file"
                self.dst_path = str(Path(self.src_path) / f"{self.name}.charm")
            self.echo(f"Building with: charm build {args}")
            ret = CharmCmd(self).build(*args.split(), _cwd=self.src_path)
        else:
            args = f"-f {self.src_path}"
            self.echo(f"Building with: charmcraft build {args}")
            ret = CharmcraftCmd(self).build(*args.split(), _cwd=self.build.build_dir)
        if not ret.ok:
            self.echo("Failed to build, aborting")
            raise SystemExit(f"Failed to build {self.name}")
```

stonepreston: I think that's much clearer

stonepreston (approved):
LGTM

PR #761, opened 2021-12-06
drop unsupported k8s versions/series; add 1.23 for upcoming release

kwmonroe:
- Drop 1.19 from our CI matrix (1.19 went [eol](https://kubernetes.io/releases/patch-releases/#non-active-branch-history) on oct 28 2021)
- Make 1.23 the default k8s version for validate jobs
  - This is probably unnecessary since everything that calls our validate jobs specifies exact versions, but it feels right to make the default accurate in case anyone uses our tests outside of our own CI.
- Drop xenial testing (16.04 went [eos](https://wiki.ubuntu.com/Releases) in april 2021)
- Add 1.23 to our test matrix


kwmonroe:
Jobs running with this PR are good -- merging.

PR #766, opened 2021-12-08
Set root disk to 80GB for MicroK8s on ARM64

ktsakalozos:
Now that we run clustering tests the arm nodes we ask from aws have to have more storage so they accomodate multiple lxc containers at a time.


neoaggelos (approved):
No idea what the current value is, would 40G be enough in our case or would it fail as well?

ktsakalozos:
@neoaggelos I think the default value is 8GB. 40GB would also be fine.

PR #770, opened 2021-12-09
Update VPC IDs in calico-spec and tigera-ee-spec

Cynerva:


Cynerva:
Merging without review to unblock 1.23 release.

PR #773, opened 2021-12-14
Sqa/keystone test optimization

addyess:
These changes address 1.23 keystone testing issues experienced with Vault + TLS Keystone

As a workaround for these two bugs, assume that the `vault's root-cert` == `keystone ca-cert` under certain circumstances
- [LP#1954835](https://bugs.launchpad.net/charm-keystone/+bug/1954835) created against keystone (ca_cert missing from `keystone-credentials` relation
- [LP#1954838](https://bugs.launchpad.net/charm-kubernetes-master/+bug/1954838) created against kubernetes-master to use `ca_cert` from relation


jobs/integration/validation.py
```diff
@@ -1372,19 +1373,56 @@ def _find_relation(*specs):
...
+                    await action.wait()
+                    assert action.status not in ("pending", "running", "failed")
+                    vault_root_ca = action.results.get("output")
+                    if vault_root_ca:
+                        vault_root_ca = base64.b64encode(
+                            vault_root_ca.encode("ascii")
+                        ).decode("ascii")
+                        break
+
+            if vault_root_ca:
```
asbalderson, line 1416: I think this is worth noting that it is a workaround for a bug 
It should make finding the code to remove when the bug is fixed easier as well

asbalderson (changes requested):
This looks pretty good to me
I left one comment for noting that setting the config via the test is a workaround for a bug.

jobs/integration/validation.py
```diff
@@ -1372,19 +1373,56 @@ def _find_relation(*specs):
...
+        credentials_rel = list(_find_relation(keystone_creds))
+        if not credentials_rel:
             await keystone.add_relation("identity-credentials", keystone_creds)
             await tools.juju_wait()
 
         keystone_master = random.choice(keystone.units)
         action = await keystone_master.run("leader-get admin_passwd")
         admin_password = action.results.get("Stdout", "").strip()
+
+        # is the same vault supplying tls-certificates for keystone, kubernetes-master, and kubernetes-worker
```
johnsca, line 1385:
```suggestion
        # Work around the following bugs which lead to the CA used by Keystone not being passed along
        # and honored from the keystone-credentials relation itself by getting the CA directly from Vault and
        # passing it in via explicit config.
        #   * https://bugs.launchpad.net/charm-keystone/+bug/1954835
        #   * https://bugs.launchpad.net/charm-kubernetes-master/+bug/1954838
```

johnsca (approved):

PR #789, opened 2022-01-21
Don't assume that 'master' is the default branch

addyess:
It's possible to fetch the default branch from github API, so rather than assume it is `master`, for charm builds, bundle builds, and sync jobs fetch the default branch if the downstream repo's branch isn't specified

### DIFF
--- a/cilib/git.py
+++ b/cilib/git.py
@@ -1,8 +1,34 @@
 """ Git utils
 """
 
+import json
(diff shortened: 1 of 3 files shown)


kwmonroe (approved):
Excellent!  This is gonna help a ton as it's so easy to lose track of which CK repos have set as the default.

PR #791, opened 2022-01-21
Charm build/branch debug

addyess:
charms `build-bundle` action uses a bundle specification that doesn't include the `donwstream` attribute.
Simplify which branch should be considered in the repo

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -239,19 +239,19 @@ class BuildEntity:
         else:
             src_path = self.checkout_path
 
-        auth = (os.environ.get("CDKBOT_GH_USR"), os.environ.get("CDKBOT_GH_PSW"))
-        default_branch = default_gh_branch(
-            opts["downstream"], ignore_errors=True, auth=auth
-        )
+        downstream = opts.get("downstream")
+        charm_branch = opts.get("branch")
 
-        if "branch" in opts:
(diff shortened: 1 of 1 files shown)


jobs/build-charms/charms.py
```diff
@@ -387,7 +387,7 @@ def _read_metadata_resources(self):
     def setup(self):
         """Setup directory for charm build"""
         downstream = f"https://github.com/{self.opts['downstream']}"
```
stonepreston, line 389: is downstream guaranteed to be set here?

addyess: TLDR: yes guaranteed.

TL part:
So this code has a lot of problems.  But `def setup(...)` is only called by the charm build job and not the bundle build-job.  The charm build uses this list [charm-support-matrix.inc](https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/includes/charm-support-matrix.inc).  
I have another MAJOR refactor (https://github.com/charmed-kubernetes/jenkins/pull/744) in this repo which addresses some of these oddities

stonepreston (commented):
LGTM, but someone with a bit more knowledge of the jenkins build process should probably take a look as well

addyess:
> LGTM, but someone with a bit more knowledge of the jenkins build process should probably take a look as well

if it helps -- someone VERY experienced (@kwmonroe) already looked at missed my mistake :)

PR #800, opened 2022-01-28
Pack bundles with `charmcraft` before publishing to charmhub

addyess:
Before publishing to charmhub, the bundle must be packed.


tests/data/cli_response/charmcraft_pack.txt
```diff
@@ -0,0 +1,2 @@
+Packing the bundle
+Created '/not/real/path/to/scratch/tmp/bundles/test-kuberenetes.zip'.
```
stonepreston, line 2:
```suggestion
Created '/not/real/path/to/scratch/tmp/bundles/test-kubernetes.zip'.
```

stonepreston (approved):
LGTM

PR #802, opened 2022-01-31
Switch over all charms publishing to charmhub rather than charmstore

addyess:
waiting on 
* decision from @ca-scribner or @knkski to decided on kubeflow bundle
  * kubeflow pieces are no longer in use.  they may be removed
* `etcd` charm to be opened for publishing on https://charmhub.io/etcd
  * this won't be available until charmstore is excluded from pushes
  * continue to push `etcd` to the charmstore
* Merge with https://github.com/charmed-kubernetes/bundle/pull/817


stonepreston (approved):
LGTM

PR #804, opened 2022-02-01
repurpose microk8s charmcraft build script for all charmcraft builds

addyess:


jobs/build-charms/charmcraft-build.sh
```diff
@@ -0,0 +1,74 @@
...
+
+_DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$_DIR" ]]; then _DIR="$PWD"; fi
+. "$_DIR/../../cilib.sh"
+
+## Requirements
+## - LXD (initialized)
+
+## Configuration
+## - REPOSITORY: Repository to pull the MicroK8s charm code from
```
stonepreston, line 11:
```suggestion
## - REPOSITORY: Repository to pull the charm code from
```

jobs/build-charms/charmcraft-build.sh
```diff
@@ -0,0 +1,74 @@
...
+fi
+
+# Configure cleanup routine
+container="${container_prefix}-${BUILD_NUMBER}"
+function cleanup() {
+  set +e
+  sudo lxc delete $container --force
+  set -e
+}
+#trap cleanup EXIT
```
addyess, line 46:
```suggestion
trap cleanup EXIT
```

addyess: oops, need this

cilib.sh
```diff
@@ -102,3 +102,13 @@ print(unit[units[0]]['public-address'])
...
     env python3 -c "import os,sys; print(os.path.dirname(os.path.abspath(\"$0\")))"
 }
+
+ci_lxc_launch()
+{
+    # Launch local LXD container to publish to charmcraft
+    local lxc_image=$1
+    local lxc_container=$2
+    sudo lxc launch ${lxc_image} ${lxc_container}
+    sleep 10
```
johnsca, line 112: Perhaps this should use the `until` loop pattern instead?

jobs/build-charms/charmcraft-build.sh
```diff
@@ -0,0 +1,74 @@
...
+ci_lxc_launch ubuntu:20.04 $container
+until sudo lxc shell $container -- bash -c 'snap install charmcraft --classic'; do
+  sleep 3
+  echo 'retrying charmcraft install'
+done
+
+
+# Build charm and fetch
+sudo lxc shell $container -- bash -c "git clone ${REPOSITORY} -b ${BRANCH} charm"
+sudo lxc shell $container --env CHARMCRAFT_MANAGED_MODE=1 -- bash -c "cd charm/$SUBDIR && charmcraft build -v"
```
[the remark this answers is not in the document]
addyess: @johnsca -- can you remind me again why `pack` is better than `build`?

i was also interested if the reactive charms we have could be built with `charmcraft pack` based on these [docs](https://discourse.charmhub.io/t/how-to-pack-your-reactive-based-charm-with-charmcraft/5452)

Is there any need for `--destructive-mode` if this runs in a temporary lxd?

cilib.sh
addyess: There was concern that the `sleep 10` was giving the container time to handle the incoming shell commands afterwards.  Without it -- `lxc shell` would hang perpetually.  The `until` would never fail and never cause a retry b/c it was blocked waiting on a result from `lxc shell` which wouldn't come back.

are you thinking

```bash
sleep 10
sudo lxc shell ${lxc_container} -- bash -c "apt-get update && apt-get install build-essential -y"
```

becomes


```bash
sleep 10
until sudo lxc shell ${lxc_container} -- bash -c 'apt-get update && apt-get install build-essential -y'; do
  sleep 3
  echo 'retrying install of build-essentials'
done
```

jobs/build-charms/charmcraft-build.sh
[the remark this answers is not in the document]
addyess: @johnsca lemme know if the change i made meets the spirit of this suggestion (it's a bit different)

cilib.sh
johnsca: Ah, right.  I was only half paying attention to that discussion.  I'm just not a fan of arbitrary sleeps, in general. If it works, that's fine.  If it becomes a problem in the future (such as on a slow runner), we can explore things like a `timeout` wrapper on the `lxc shell` or something.

johnsca (approved):
LGTM

neoaggelos (approved):

PR #806, opened 2022-02-02
Charmhub/charmcraft build in lxd

addyess:
* ensure that before running `charmcraft pack` every charm has a valid `version`.  
    * `charmcraft` doesn't do this automatically for operator charms https://github.com/canonical/charmcraft/issues/37
* rename `legacy_charm` to `reactive` and remove old code trying to deal with non-reactive charms that were bare hook charms
* by default, assume a `.charm` file is built (only charmstore reactive charms do not require this)
* identify charm versions by git commit hashes (short hash for charmhub, long hash for charmstore, and .build.manifest for reactive charms).

There's quite a bit of work involved around deciding if a charm should be rebuilt or not based on comparing the charm's local git hash and the posted charm's git hash. There's a matrix of how to determine which hashes to consider:

|    | reactive | hash-type | 
| --- | --- | --- |
| **charmhub** | true | pull released charm, unzip, compare `.bundle.manifest` with local layers |
| **charmstore** | true | pull released charm, unzip, compare `.bundle.manifest` with local layers |
| **charmhub** | false | short hash (eg `eb7a891e`) |
| **charmstore** | false | long hash (eg `eb7a891e649c34b90560e4bd4ef5bbcfcac2261e`) |


stonepreston (approved):
LGTM

johnsca (approved):
Looks reasonable to me.

PR #807, opened 2022-02-03
Optimize multiple charms being built by the same charmcraft container

addyess:
Rather than every charm being built in its own separate container, reuse the same container for every charmbuild, careful to clean up the working space before each build. 

Refactor so the origin build script `charmcraft-build.sh` can continue to be used by `build-microk8s-charm` job

### DIFF
--- a/jobs/build-charms.yaml
+++ b/jobs/build-charms.yaml
@@ -105,6 +105,8 @@
             #!/bin/bash
             set -eux
 
+            . $WORKSPACE/cilib.sh
+            
             sudo chown -R jenkins:jenkins /var/lib/jenkins/.config/ || true
 
             IS_FORCE=""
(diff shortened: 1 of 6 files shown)


jobs/build-charms.yaml
```diff
@@ -114,6 +116,14 @@
...
             rm -rf "$WORKSPACE/charms" || true
 
+            # Cleanup old charmcraft containers
+            ci_lxc_delete "${JOB_NAME}"
+            
+            # Configure cleanup routine for exit
+            charmcraft_lxc="${JOB_NAME}-${BUILD_NUMBER}"
+            trap 'ci_lxc_delete $charmcraft_lxc' EXIT
+            ci_charmcraft_launch $charmcraft_lxc
+
```
addyess, line 126: startup the `charmcraft_lxc` build container

addyess:
@neoaggelos, i believe the microk8s charm should still build with the build script.  It continues to work on my machine locally

stonepreston (commented):
LGTM

kwmonroe (approved):
LGTM

PR #811, opened 2022-02-09
remove the charm login task

kwmonroe:
The store has transitioned over to charmhub.  There is no need to have a login task for infra jobs because we use the hub creds envvar any time we need to interact with charmhub.


addyess (approved):
TOTALLY!

PR #814, opened 2022-02-14
use default_branch on sync-charms job rather than `master`

addyess:


cilib/service/charm.py
```diff
@@ -30,13 +30,15 @@ def sync(self):
             return
 
         self.log(f"Syncing {self.upstream_normalized} -> {self.downstream_normalized}")
+        upstream_ref = self.repo.default_gh_branch("upstream")
+        downstream_ref = self.repo.default_gh_branch("downstream")
```
kwmonroe, line 34:
```suggestion
        upstream_ref = self.repo.default_gh_branch(self.upstream_normalized)
        downstream_ref = self.repo.default_gh_branch(self.downstream_normalized)
```

These need to be repo url strings, right?  Or perhaps your intention was to make `repo.default_gh_branch` handle the literal "upstream" vs "downstream" as a param?

cilib/service/charm.py
```diff
@@ -30,13 +30,15 @@ def sync(self):
...
             src_path = Path(tmpdir) / self.repo.src
             self.repo.base.clone(cwd=tmpdir)
             self.repo.base.remote_add(
                 origin="upstream", url=self.repo.upstream, cwd=str(src_path)
             )
             self.repo.base.fetch(origin="upstream", cwd=str(src_path))
-            self.repo.base.checkout(ref="master", cwd=str(src_path))
-            self.repo.base.merge(origin="upstream", ref="master", cwd=str(src_path))
-            self.repo.base.push(origin="origin", ref="master", cwd=str(src_path))
+            self.repo.base.checkout(ref=upstream_ref, cwd=str(src_path))
```
kwmonroe, line 42:
```suggestion
            self.repo.base.checkout(ref=downstream_ref, cwd=str(src_path))
```

We're in the downstream repo at this point, so we want to checkout the downstream default branch.

tests/unit/build_charms/conftest.py
```diff
@@ -0,0 +1,16 @@
+"""Ensure that path is included in PYTHONPATH before tests are run."""
+import importlib
+import sys
+import pytest
+
+
+@pytest.fixture(scope="package")
+def charms():
```
kwmonroe, line 8: :heart:

kwmonroe (changes requested):
Nice job on the unit tests (even if you did sneak in a build-charms update :)

I think the default_gh_branch and checkout bits need tweaking.

tests/unit/build_charms/conftest.py
addyess: this "sneak in" was part of adding new unit tests.  because the build python are scattered around, they aren't in python packages -- they're just bare python files.  This means the tests have to open/close scope on the python path.  "charms.py" in one directory isn't necessarily the same module as "charms.py" in other path.

kwmonroe: Makes sense, thanks!

kwmonroe (approved):

PR #816, opened 2022-02-15
Fail charm builds only after attempting to build each one

addyess:
Rather than raise a `SystemExit` which bypasses any exception handling, use `BuildException` which can be caught.  After all the charms are built -- raise the last exception to cease the build process.

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -680,7 +680,7 @@ class BuildEntity:
             echo=self.echo,
         )
         if not ret.ok:
-            raise SystemExit("Clone failed")
+            raise BuildException("Clone failed")
 
         self.reactive = self.layer_path.exists()
 
@@ -719,7 +719,7 @@ class BuildEntity:
 
         if not ret.ok:
             self.echo("Failed to build, aborting")
-            raise SystemExit(f"Failed to build {self.name}")
+            raise BuildException(f"Failed to build {self.name}")
 
     def push(self):
         """Pushes a built charm to Charm store/hub."""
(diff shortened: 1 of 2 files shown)


jobs/build-charms/charms.py
```diff
@@ -974,9 +976,16 @@ def build(
             entity.push()
             entity.attach_resources()
             entity.promote(to_channel=to_channel)
+        except Exception as build_exc:
+            failed_entities.append((entity, build_exc))
```
Cynerva, line 980: I recommend something like `entity.echo(traceback.format_exc())` here so we get tracebacks for all the exceptions, not just the last one.

jobs/build-charms/charms.py
```diff
@@ -974,9 +977,15 @@ def build(
...
             entity.attach_resources()
             entity.promote(to_channel=to_channel)
+        except Exception:
+            entity.echo(traceback.format_exc())
+            failed_entities.append(entity)
         finally:
             entity.echo("Stopping")
 
+    if any(failed_entities):
+        raise BuildException("Encountered Charm Build Failure")
```
kwmonroe, line 987: We're not doing anything with the list of `failed_entities`.  Perhaps add a `len()` to this output so we know how many charms failed?  Or swap the list for a simple bool flag that tells us if at least one thing failed?

kwmonroe (approved):
Love the spirit of this!  Minor knit that we're not doing anything with our curated list of failures, but i'm +1 as is.

jobs/build-charms/charms.py
addyess: i'm on it

addyess:
```bash
[canal] Stopping
Encountered 1 Charm Build Failure:
	canal
```

PR #817, opened 2022-02-16
ignore errors in default_gh_branch (not all layers comes from gh)

kwmonroe:
Follow-on to #814.

Not all our layer repos come from GitHub (e.g. `layer-apt`, `layer-snap`, etc).  Ignore errors in `default_gh_branch()`.  If we can't figure out a default branch name, set it to `master`.


addyess (approved):
Yep, this is what `ignore_errors` is for!

PR #818, opened 2022-02-16
Remove calico-upgrade resource from canal charm

Cynerva:
I believe this is needed to go with https://github.com/charmed-kubernetes/layer-canal/pull/67


kwmonroe (approved):
Nice catch!

PR #825, opened 2022-02-18
juju 2.9.7 (starting in 2.9.4) no longer needs the async event loops passed to Model()

addyess:
This resulted in a breaking change in our integration tests

Removing the event_loop from the model, showed the module level fixture was no longer needed.  But `event_loop` is a now a function based pytest fixture as a part of `pytest-asyncio`.  

By setting `asyncio_mode = auto` in pytest.ini, all fixtures can now `await` things automatically.

Here's the [breaking change in libjuju](https://github.com/juju/python-libjuju/commit/e5b1993eba0e466a38d593732bf62df8f90739c3)

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -164,8 +164,8 @@ async def tools(request):
 
 
 @pytest.fixture(scope="module")
-async def model(request, event_loop, tools):
-    model = Model(event_loop)
+async def model(request, tools):
+    model = Model()
     await model.connect(tools.connection)
     if request.config.getoption("--is-upgrade"):
         upgrade_snap_channel = request.config.getoption("--upgrade-snap-channel")
@@ -232,7 +232,7 @@ async def k8s_model(model, tools):
             "--no-switch",
         )
         created_k8s_model = True
-        k8s_model = Model(model.loop)
+        k8s_model = Model()
         await k8s_model.connect(tools.k8s_connection)
         yield k8s_model
     finally:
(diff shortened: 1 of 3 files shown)


jobs/integration/conftest.py
```diff
@@ -164,8 +164,8 @@ async def tools(request):
 
 
 @pytest.fixture(scope="module")
-async def model(request, event_loop, tools):
-    model = Model(event_loop)
+async def model(request, tools):
+    model = Model()
```
addyess, line 168: Here was the main problem that was causing errors in SOL-QA

jobs/integration/conftest.py
```diff
@@ -232,7 +232,7 @@ async def k8s_model(model, tools):
             "--no-switch",
         )
         created_k8s_model = True
-        k8s_model = Model(model.loop)
+        k8s_model = Model()
```
addyess, line 235: This exposed the same problem.  Starting in juju 2.9.5, Model's first argument was no longer an event_loop, but instead `max_frame_size`

jobs/integration/conftest.py
```diff
@@ -399,10 +399,11 @@ def log_dir(request):
     return path
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def event_loop():
     loop = asyncio.new_event_loop()
     yield loop
+    loop.close()
```
addyess, line 406: overrides the fixture in `pytest-asyncio:event_loop` which is a "function" based scope to instead be session based so that modules based fixtures can be `async/await`

PR #831, opened 2022-02-18
upload eks snaps from within the build container

kwmonroe:
Jenkins has snap creds for k8s and snap creds for cpc.  If build jobs for both happen to overlap, one of those jobs is doomed.  That's because we do things like `snapcraft [login|logout]` at the jenkins worker layer, which potentially affects snap store auth/tokens for all concurrent jobs on that worker.

This PR mitigates that by moving the snapcraft bits that require authn *inside* the isolated build container.

Drive-by to ensure we at least have *some* `base:` declaration; recent snapcraft versions won't build without it.

### DIFF
--- a/jobs/build-snaps/build-release-eks-snaps.groovy
+++ b/jobs/build-snaps/build-release-eks-snaps.groovy
@@ -22,13 +22,13 @@ pipeline {
         timestamps()
     }
     stages {
-        stage('Setup User') {
+        stage('Verify User') {
             /* Do this early; we want to fail fast if we don't have valid creds. */
             steps {
                 sh """
-                    snapcraft logout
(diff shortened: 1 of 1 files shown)


jobs/build-snaps/build-release-eks-snaps.groovy
```diff
@@ -56,6 +56,12 @@ pipeline {
                         cd \${EKS_SNAP}
                         sed -i -e "s/^name: \${snap}/name: \${EKS_SNAP}/" \
                                -e "s/^base: .*/base: ${EKS_BASE}/" snapcraft.yaml
+
+                        # if we don't have any base defined at this point, add one
+                        grep -q "^base: " snapcraft.yaml || echo "base: ${EKS_BASE}" >> snapcraft.yaml
```
addyess, line 61: a little on the fly bash editing of yaml.  Gotta love it

PR #835, opened 2022-02-25
update cdk-addons jobs

kwmonroe:
EKS snap builds are going smashingly well.   cdk-addons snap builds fail sometimes because when multiple snap-related jobs happen on the same jenkins worker, one `snapcraft logout` can affect the auth for all other jobs.

EKS doesn't have this problem because that job builds and uploads from within a lxc container.  This PR does the same for cdk-addons.

- update job description and supported versions
- drop arch param (all our arches are built on amd64)
- build snaps in container like we do for eks jobs
- upload snaps from within the build container

### DIFF
--- a/jobs/build-snaps/build-release-cdk-addons.groovy
+++ b/jobs/build-snaps/build-release-cdk-addons.groovy
@@ -253,6 +253,10 @@ pipeline {
                             sudo lxc shell ${lxc_name} -- bash -c "snapcraft login --with /snapcraft-creds"
                             for arch in ${env.ADDONS_ARCHES}
                             do
+                                if [ "\${arch}" = "ppc64le" ]
+                                then
+                                    arch="ppc64el"
+                                fi
                                 BUILT_SNAP="cdk-addons_${kube_ersion}_\${arch}.snap"
 
                                 echo "Uploading \${BUILT_SNAP}."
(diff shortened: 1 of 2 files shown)


jobs/build-snaps/build-release-cdk-addons.groovy
```diff
@@ -88,76 +86,59 @@ pipeline {
...
-                          SNAPCRAFT_BUILD_ENVIRONMENT=host snapcraft --target-arch=\${arch} || exit 1
-                          mv *.snap ..
-                          cd ..
+                          arch="ppc64el"
                         fi
-                    done
-                    cd ..
+                        sed -i "s/KUBE_ARCH/\${arch}/g" build/snapcraft.yaml
 
-                    echo "Updating bundle with upstream images."
```
kwmonroe, line 126: The purpose of this job is to pull CK images from our staging area, then tag/push them to our production area. There is no need to manipulate the bundle's `container-images.txt` file in this job because it already (rightfully) happens in the [sync-oci-images](https://github.com/charmed-kubernetes/jenkins/blob/98a89b0166a9a94211a1c8804c6e65192f667c53/jobs/sync-oci-images/sync-oci-images.groovy#L75-L108) job.

kwmonroe:
This looks good:

https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-1.22/5/

PR #837, opened 2022-02-25
use cilib to manage build container; pull a reasonable jujud-op image

kwmonroe:
We need the [jujud-operator image](https://hub.docker.com/r/jujusolutions/jujud-operator/tags) in rocks so we can bootstrap k8s models on vsphere, but the image tag has the juju version baked in. It's painful to keep updating https://github.com/charmed-kubernetes/bundle/blob/master/container-images.txt every time the juju client  gets a new version.

This PR determines the version of the juju snap installed on the jenkins worker, constructs a jujud-operator image string, and injects that for processing by our `Process CI Images` stage.

It's not perfect since we can't guarantee that our whole team is using the same juju version as the jenkins worker, but it's better than the current drift between new juju snaps and a hard-coded value in `container-images.txt`.

### DIFF
--- a/jobs/sync-oci-images/sync-oci-images.groovy
+++ b/jobs/sync-oci-images/sync-oci-images.groovy
@@ -7,10 +7,7 @@ def kube_ersion = null
 if (kube_version != "") {
     kube_ersion = kube_version.substring(1)
 }
-def lxc_name = 'sync-image-processor'
-def lxd_exec(String container, String cmd) {
-    sh "sudo lxc exec ${container} -- bash -c '${cmd}'"
-}
+def lxc_name = env.JOB_NAME+"-"+env.BUILD_NUMBER
 
 pipeline {
     agent {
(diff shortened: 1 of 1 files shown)


kwmonroe:
This looks good:

https://jenkins.canonical.com/k8s/job/sync-oci-images/396/

jobs/sync-oci-images/sync-oci-images.groovy
```diff
@@ -292,12 +302,17 @@ pipeline {
     }
     post {
         always {
-            echo "All images known to this builder:"
-            sh "sudo lxc exec ${lxc_name} -- ctr image ls"
```
kwmonroe, line 296: we delete images after every push, so this doesn't do anything useful.

addyess (approved):
This is clever.  So the jenkins worker gets it snaps updated, and we pick and choose the jujud image id number from the version in our local snap

addyess:
👍🏻 to merge when you're ready

PR #839, opened 2022-02-28
sub . for - in the job name so lxc can make a uniquely named container

kwmonroe:
I noticed cdk-addons 1.21 and 1.24 build at roughly the same time each night. If they overlap, one job will delete the lxc build container before the other job completes.

Fix this by making the lxc name unique.  Note, lxc doesn't like dots in the container name, and these jobs are named `foo-1.2x`; replace those dots with dashes.

Successful run from this branch:

https://jenkins.canonical.com/k8s/job/build-release-cdk-addons-1.21/5/
```
...
09:23:20  + sudo lxc launch ubuntu:20.04 build-release-cdk-addons-1-21-5
09:23:20  Creating build-release-cdk-addons-1-21-5
...
```

### DIFF
--- a/jobs/build-snaps/build-release-cdk-addons.groovy
+++ b/jobs/build-snaps/build-release-cdk-addons.groovy
@@ -7,7 +7,7 @@ def kube_ersion = null
 if (kube_version != "") {
     kube_ersion = kube_version.substring(1)
 }
-def lxc_name = "build-release-cdk-addons-"+env.BUILD_NUMBER
+def lxc_name = env.JOB_NAME.replaceAll('\\.', '-')+"-"+env.BUILD_NUMBER
 
 pipeline {
     agent {


jobs/build-snaps/build-release-cdk-addons.groovy
```diff
@@ -7,7 +7,7 @@ def kube_ersion = null
 if (kube_version != "") {
     kube_ersion = kube_version.substring(1)
 }
-def lxc_name = "build-release-cdk-addons-"+env.BUILD_NUMBER
+def lxc_name = env.JOB_NAME.replaceAll('\\.', '-')+"-"+env.BUILD_NUMBER
```
kwmonroe, line 10: groovy `replaceAll` takes a regex as the first arg, so we need a backslash to escape the backslash that escapes the dot.

jobs/build-snaps/build-release-cdk-addons.groovy
```diff
@@ -7,7 +7,7 @@ def kube_ersion = null
 if (kube_version != "") {
     kube_ersion = kube_version.substring(1)
```
addyess, line 8: `kube_ersion`?  Where's the `v`

jobs/build-snaps/build-release-cdk-addons.groovy
addyess: gotcha. container names don't like `.`.  So this changes `.` to `-` using regex.

jobs/build-snaps/build-release-cdk-addons.groovy
addyess: can we fix this in a drive by PR?

Cynerva: It's by design:

```
kube_version=v1.23.0
kube_ersion=1.23.0
```

I will be sad the day kube_ersion goes away because it amuses me greatly. But do as you will.

addyess: OH MY. Please leave it because it's SO clear

wwwtyro: LGTM

PR #848, opened 2022-03-22
Fix skipped builds when local version matches latest release version

Cynerva:
Some charm builds are being skipped because the local version is being compared against the remote version of the latest release. These builds still need to happen so they can be pushed to the destination channel.

This PR fixes it by comparing the local version against the remote version of the destination channel, instead.


kwmonroe (approved):
Blackened and +1.

PR #849, opened 2022-03-23
refresh snapd snap along with core in validate-snapd-upgrade

kwmonroe:
The snapcore team periodically asks us to test new beta `core` snaps prior to them going stable.  We'd been doing this for a while, but recently questioned the value of that test considering all our k8s snaps are now based on `core18`.

We mentioned this to the team and got the following response:

> Now your snaps are interacting with snapd snap in almost the same way they were interacting with core snap before. So that is why I suggested testing when snapd snap goes to beta.

This PR refreshes both the `core` and `snapd` snaps during `validate-ck-snapd-upgrade` so we once again ensure new snapd/core snaps don't affect CK.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -77,7 +77,7 @@ def pytest_addoption(parser):
         help="Charm channel to use (stable, candidate, beta, edge)",
     )
 
-    # Set when testing a different snap core channel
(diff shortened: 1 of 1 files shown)


Cynerva (approved):

Cynerva:
:+1: Thanks for this.

PR #855, opened 2022-03-24
adjust release-microk8s to pipeline job

addyess:


kwmonroe (approved):
Wow, really nice job.

TIL you can nest a `script` under a `stages->stage->steps->script->each->stages->stage->steps` block and not hit any turtles.

I haven't seen this run in CI, but given that we broke `release-microk8s` by necessity in #853, it can't get any worse. Merge at will.

addyess:
@ktsakalozos 
* this should address removing the OGC dependancy on release-microk8s-{arch} jobs
* the options for an arch specific job are "all, {arch}/beta, {arch}/stable, {arch}/pre-release"
   *  this changes from before where you all would run unrelated arches.  Have i done this correctly?
* this creates a new jenkins job called `release-microk8s-arch-{arch}`.  
    * If this name isn't preferred, we should delete the freestyle jenkins job named `release-microk8s-{arch}` and use `jjb` to update the job definitions

jobs/release-microk8s/release-microk8s.groovy
```diff
@@ -0,0 +1,131 @@
...
+                                    --model-default test-mode=true \
+                                    --model-default resource-tags=owner=k8sci \
+                                    --bootstrap-constraints "instance-type=${instance_type}"
+
+                                juju deploy -m "${juju_full_model}" --constraints "${constraints}" ubuntu
+
+                                juju-wait -e "${juju_full_model}" -w
+
+                                juju ssh -m "${juju_full_model}" --pty=true ubuntu/0 -- 'sudo snap install lxd'
+                                juju ssh -m "${juju_full_model}" --pty=true ubuntu/0 -- 'sudo lxd.migrate -yes' || true
```
addyess, line 98: previous ogs scripts were running with `set +e` which would  not error on bash errors, whereas jenkins runs with `set -e` be default.  this line seems to fail on jenkins every time but I don't have old runs to see if it previously failed or not.  That's why there's a `|| true` at the end of this command

jobs/release-microk8s/release-microk8s.groovy
```diff
@@ -0,0 +1,131 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+    post {
+        always {
+            script {
+                all_tracks.each { track ->
```
addyess, line 124: this post is here in case any previous controllers weren't successfully cleaned up

jobs/release-microk8s/release-microk8s.groovy
```diff
@@ -0,0 +1,131 @@
...
+        stage("Snapcraft login") {
+            steps {
+                sh "snapcraft login --with ${SNAPCRAFTCREDS}"
+            }
+        }
+        stage("Setup tox environment") {
+            steps {
+                sh """
+                tox -e py38 -- python -c 'print("Tox Environment Ready")'
+                """
```
addyess, line 58: this setups up the tox environment once, so every track run doesn't need it re-invoked

jobs/release-microk8s/release-microk8s.groovy
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+        stage("Release Tracks") {
+            steps {
+                script {
+                    all_tracks.each { track -> 
+                        stage("Track ${track}") {
+                            script {
```
addyess, line 66: this script was ported from the deleted `spec.yaml`

jobs/release-microk8s.yaml
```diff
@@ -33,26 +49,9 @@
             Choose a specific architecture and track to release against
           choices:
             - all
-            - amd64/beta
-            - arm64/beta
-            - amd64/stable
-            - arm64/stable
-            - amd64/prerelease
-            - arm64/prerelease
```
addyess, line 41: i'm not **SURE** but i think running 'all' previously would have tried to build all 3 tracks on both archs.  This has been scaled back since there are two jobs (one per arch) which will build all 3 tracks within that arch

PR #871, opened 2022-04-06
add k-n-b layer

kwmonroe:
Add the new `kubernetes-node-base` layer. Once we're sure we'll never have to build charms for < 1.24, we can remove the `kubernetes-master-worker-base` stanza.


addyess (approved):
🚀 💯 LGTM

PR #881, opened 2022-04-26
Validation fixes

kwmonroe:
Couple validation fixes:

- pick a better charm from which to gleen unit series for the series-upgrade test
- make the snapd-upgrade job more like others (using `ci::run`)
- let microbot run on vsphere (no reason for that test to be ec2-only)
- don't use the ceph-admin relation any more (k-c-p has been refactored to only use ceph-client)


kwmonroe:
Looks good per 4/26 [ci report](http://jenkaas.s3-website-us-east-1.amazonaws.com/).

PR #888, opened 2022-04-28
Fixes for the release-microk8s-charm job

neoaggelos:
## Summary

Update the release-microk8s-charm job.

## Changes

- Run tests for focal and jammy (needs https://github.com/canonical/charm-microk8s/pull/35)
- Use charmcraft credentials properly
- Use public charmhub API

### DIFF
--- a/jobs/microk8s/charms/release.sh
+++ b/jobs/microk8s/charms/release.sh
@@ -7,72 +7,57 @@
 ## Configuration
 ## - REPOSITORY: Repository to pull the MicroK8s charm code from
 ## - TAG: Tag to checkout
-## - CHARM_NAME: Charm name to test (microk8s)
-## - FROM_CHANNEL: Pull charm from this channel (latest/edge)
-## - TO_CHANNEL: After running tests, release revision to this channel (latest/stable)
-## - SKIP_TESTS: Do not run tests
-## - SKIP_RELEASE: Do not release
+## - CHARM_NAME: Name of MicroK8s charm (default: microk8s)
+## - FROM_CHANNEL: Run integration tests against this channel. (default: edge)
+## - TO_CHANNEL: After tests pass, charm will be pushed to this channel. (default: stable)
+## - CLUSTER_SIZE: Cluster size for the integration tests.
+## - SNAP_CHANNELS: MicroK8s snap channels to test (space separated list).
+## - SERIES: Run tests for these OS series (space separated list) (default: focal jammy).
+## - SKIP_TESTS: Skip tests for promoting release (NOT RECOMMENDED).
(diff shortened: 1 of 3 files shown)


jobs/microk8s/charms/release.sh
```diff
@@ -7,72 +7,57 @@
...
-# {
-#   "channel": "latest/stable",
-#   "revision": 3
-# }
-revision=$(sudo lxc shell $container -- \
-  bash -c "curl https://api.charmhub.io/v1/charm/$CHARM_NAME/releases -b ~/snap/charmcraft/common/config/charmcraft.credentials" \
-    | jq -r ".[\"channel-map\"][] | select(.channel == \"$FROM_CHANNEL\") | .revision")
+# curl 'https://api.charmhub.io/v2/charms/info/microk8s?fields=channel-map' | jq -r '.["channel-map"][] | select(.channel.name == "edge") | select(.channel.base.architecture == "amd64") | .revision.revision' | head -1
+# 20
+revision=$(curl --silent 'https://api.charmhub.io/v2/charms/info/microk8s?fields=channel-map' | jq -r '.["channel-map"][] | select(.channel.name == "'"$FROM_CHANNEL"'") | select(.channel.base.architecture == "amd64") | .revision.revision' | head -1)
```
addyess, line 56: this looks like fun.  i'm guessing there's not a `charmcraft` command for this?

something like [this](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/build-charms/charms.py#L218-L222)

```bash
$ charmcraft revisions microk8s
```

isn't fun to parse the table'd output

addyess (approved):
nothing is JUMPING out to me right away.  

Different question: is there a reason the microk8s charm isn't built with the [charm-support-matrix](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/includes/charm-support-matrix.inc)?

kwmonroe (approved):

jobs/microk8s/charms/release.sh
neoaggelos: > isn't fun to parse the table'd output

if it had a parsable output option, i would be much happier. it's challenging because of the multiple lines and architectures getting in the way.

addyess: https://github.com/canonical/charmcraft/issues/183

neoaggelos:
> Different question: is there a reason the microk8s charm isn't built with the [charm-support-matrix](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/includes/charm-support-matrix.inc)?

We should probably move to that at some point. The job predates the charmhub support of charm-support-matrix.

PR #892, opened 2022-05-02
adjust descriptions and values for version vars for the 1.24 GA

kwmonroe:
Upstream is set to GA 1.24 tomorrow. Before that happens, we need to adjust version-related enums to stop checking for 1.24 pre-releases.

Also update descriptions of these enums to hopefully make it more clear *when* these should be changed.

### DIFF
--- a/cilib/enums.py
+++ b/cilib/enums.py
@@ -5,20 +5,22 @@ import yaml
 
 JOBS_PATH = Path("jobs")
 
-# Current supported STABLE K8s MAJOR.MINOR release
-# This should be updated whenever a new CK major.minor is GA
+# Current supported STABLE K8s MAJOR.MINOR release. This determines what the
+# latest/stable channel is set to. It should be updated whenever a new CK
+# major.minor is GA.
 K8S_STABLE_VERSION = "1.23"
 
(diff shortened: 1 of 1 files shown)


cilib/enums.py
```diff
@@ -54,7 +56,7 @@
     "1.21": ["1.21/stable", "1.21/candidate", "1.21/beta", "1.21/edge"],
     "1.22": ["1.22/stable", "1.22/candidate", "1.22/beta", "1.22/edge"],
     "1.23": ["1.23/stable", "1.23/candidate", "1.23/beta", "1.23/edge"],
-    "1.24": ["1.24/candidate", "1.24/beta", "1.24/edge"],
+    "1.24": ["1.24/stable", "1.24/candidate", "1.24/beta", "1.24/edge"],
```
kwmonroe, line 59: Note these are the *snap* tracks, not charm tracks. It's reasonable to promote the GA snap to `1.24/stable` even if the CK GA is still being ironed out.

addyess (approved):
i don't see anything wrong, and i appreciate the extra details.  I suppose a thorough review would include me looking where these variables are used

PR #896, opened 2022-05-03
check if the DRY_RUN param is 'true'

addyess:


addyess:
we have to check if `DRY_RUN` is "true", not non-empty.  

because DRY_RUN="false" is non-empty which would enable `--dry-run`

stonepreston (approved):

PR #897, opened 2022-05-04
re-tag charms that ride on a kubernetes provider cloud, rather than a…

addyess:
I'm not sure the name `k8s-operator` best describes these charms, but they aren't machine charms and aren't dependent on a specific version of kubernetes necessarily to operate.  Unlike the `k8s` tagged charms which are charms associated with the operation of charmed-kubernetes


kwmonroe (approved):
LGTM, i'm +1 to the `k8s-operator` name.

PR #898, opened 2022-05-04
poison CHARM_BASE_DIR path from responding to git rev-parse

addyess:
[here's](https://github.com/juju/charm-tools/blob/master/charmtools/build/builder.py#L277-L291) where charmtools builds the layers list
[here's](https://github.com/juju/charm-tools/blob/master/charmtools/build/builder.py#L105-L108) where it's using a local directory
[here's](https://github.com/juju/charm-tools/blob/6ff9c20032f140771064fef6f655b655a33efb8c/charmtools/fetchers.py#L110-L127) where it pulls the rev
and because the layers and interfaces are pulled into the workspace's folder, and there's no `.git` folder there -- it searches up til it hits a `git` folder -- you end up with hit on the jenkins workspace path which provides its `git rev-parse HEAD`

This PR poisons the `git rev-parse` command to fail by putting a `.git` file in `$WORKSPACE/.cache/charmbuild/$BUILD_TAG`

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -316,6 +316,7 @@ class BuildEnv:
     def __new__(cls, *args, **kwargs):
         """Initialize class variables used during the build from the CI environment."""
         try:
+            cls.base_dir = Path(os.environ.get("CHARM_BASE_DIR"))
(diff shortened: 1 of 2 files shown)


stonepreston (commented):
LGTM

stonepreston (approved):

PR #912, opened 2022-05-12
new ps5 workers are online, lets maintain em

kwmonroe:
In preparation of the migration to PS5, IS added 4 new workers to our jenkaas env.  Add new infra jobs for these so we can put them to use.  We'll keep these separate from the old `infra-*` jobs so we can easily kill those once the migration is complete.

https://jenkins.canonical.com/k8s/view/Infra/


kwmonroe:
`ps5-amd64-[8|9|11|12]` queued up and ran successfully.

PR #913, opened 2022-05-12
report with current validate naming scheme

kwmonroe:
We now split validate jobs by cloud (aws / vsphere).  Drop the old entries (`validate-ck-amd64`) from the report since we have `validate-ck-$cloud-$arch` entries now.


kwmonroe:
http://jenkaas.s3-website-us-east-1.amazonaws.com/

PR #936, opened 2022-06-13
Charmstore removal

addyess:
Changes to remove ever pushing to charmstore again since uploads have been shutdown


addyess:
These changes will need `jjb` to update some of the jobs

kwmonroe (approved):
LGTM.  it's just `build-charms.yaml` (with new `ci-master`) that needs to be re-run through jjb, right?

kwmonroe:
> Yes, i believe that's true because i removed a `--store` parameter from the `click` invocation

Cool, job refreshed.

PR #940, opened 2022-06-18
update arc conformance to use containerized testsuite

kwmonroe:
The [arc test harness](https://github.com/Azure/azure-arc-validation) is now containerized.

- Update our job to support this
- Azure's default VM cidr overlaps our default calico cidr; ensure these are separate
- Use the k8s-apiserver semver in our product metadata


addyess (approved):
LGTM.  let's get rid of those old wheels

PR #947, opened 2022-06-27
specify that the vsphere-cloud-provider shouldn't be pushed to any charmhub track less than 1.25

addyess:
Implement range boundaries for charmhub tracks and risks to prevent charms from landing in tracks which shouldn't be opened

Required a new class called `ChannelRange`

```python
assert "latest/edge" in ChannelRange("1.18", "1.25/stable") # latest/<anything> is ignored
assert "1.24/edge" in ChannelRange("1.18", "1.25/stable")   # within bounds inclusively
assert "1.24/edge" in ChannelRange("1.18", None)            # No upper bound
assert "1.24/edge" in ChannelRange(None, "1.25/stable")     # No lower bound
assert "1.24/edge" in ChannelRange(None, None)              # No bounds
```

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -28,9 +28,10 @@ from cilib.enums import SNAP_K8S_TRACK_MAP
 from cilib.service.aws import Store
 from cilib.run import cmd_ok, capture, script
 from datetime import datetime
+from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


kwmonroe (approved):
This is really clever (and i'm stealing `latest/banana` for all my unit tests).  +1

PR #955, opened 2022-07-14
it seems snapcraft upload command doesn't accept -d as an argument 

addyess:
Replace that argument with `-v`


stonepreston (approved):
LGTM

PR #968, opened 2022-08-01
build charm bundles even when charm builds fail

addyess:
use `bash` variable to collect the EXIT_STATUS of each build command, and fail only when either or both fail


addyess:
I believe this will need `jjb` to update the job and can be done anytime before actually merging...

kwmonroe (approved):

kwmonroe:
jjb has been run on `jobs/build-charms.yaml`

PR #971, opened 2022-08-03
Adjust promote method to only promote one charm revision.

addyess:
With the introduction of charmhub bases, multiple charm revisions could end up matching some specific `from_channel` like say for instance `latest/candidate`

There could be more than 1 `latest/candidate` release from days gone by on a base (LTS + ARCH) which is now no longer supported by the charm.

For the time being until we make better use of charm tracks, bases, and charmhub in general it's reasonable to assume that the most recent charm version in a `from_channel` is the one that was intended to be promoted.

### DIFF
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -352,12 +352,7 @@ class _CharmHub(CharmcraftCmd):
             charm_status = [
                 row
                 for row in self.status(charm_entity)
-                if all(
-                    (
-                        row["Revision"],
-                        f"{row['Track']}/{row['Channel']}" == from_channel,
-                    )
-                )
+                if row["Revision"] and f"{row['Track']}/{row['Channel']}" == from_channel
             ]
 
         calls = set()
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -352,7 +352,8 @@ class _CharmHub(CharmcraftCmd):
             charm_status = [
                 row
                 for row in self.status(charm_entity)
-                if row["Revision"] and f"{row['Track']}/{row['Channel']}" == from_channel
+                if row["Revision"]
+                and f"{row['Track']}/{row['Channel']}" == from_channel
             ]
 
         calls = set()
--- a/jobs/build-charms/charms.py
+++ b/jobs/build-charms/charms.py
@@ -245,6 +245,7 @@ class CharmcraftCmd(_WrappedCmd):
 
 class _CharmHub(CharmcraftCmd):
     STATUS_RESOURCE = re.compile(r"(\S+) \(r(\d+)\)")
+    BASE_RE = re.compile(r"(\S+) (\S+) \((\S+)\)")
 
     @staticmethod
     def _table_to_list(header, body):
@@ -269,18 +270,24 @@ class _CharmHub(CharmcraftCmd):
 
     def status(self, charm_entity):
         """Read CLI Table output from charmcraft status and parse."""
-        charm_status = self.charmcraft.status(charm_entity)
+        charm_status = self.charmcraft.status(charm_entity, _out=None)
         header, *body = charm_status.stdout.decode().splitlines()
         channel_status = self._table_to_list(header, body)
 
         for idx, row in enumerate(channel_status):
-            resources = row.get("Resources", "")
-            if resources == "-":
-                row["Resources"] = []
-            elif resources == "↑":
-                row["Resources"] = channel_status[idx - 1]["Resources"]
-            else:
-                row["Resources"] = dict(self.STATUS_RESOURCE.findall(resources))
+            base_split = self.BASE_RE.findall(row.get("Base"))
+            assert base_split, "Failed to split base into name, channel, and arch"
+            row["Base"] = dict(zip(["name", "channel", "architecture"], base_split[0]))
+            for prop in ["Resources", "Revision", "Version"]:
+                value = row.get(prop, "")
+                if value == "↑":
+                    row[prop] = channel_status[idx - 1][prop]
+                elif value == "-":
+                    row[prop] = {} if prop == "Resources" else None
+                elif prop == "Resources":
+                    row[prop] = dict(self.STATUS_RESOURCE.findall(value))
+                else:
+                    row[prop] = value
         return channel_status
 
     def revisions(self, charm_entity):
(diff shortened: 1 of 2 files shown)


jobs/build-charms/charms.py
```diff
@@ -269,18 +270,24 @@ def info(name, **query):
...
-            resources = row.get("Resources", "")
-            if resources == "-":
-                row["Resources"] = []
-            elif resources == "↑":
-                row["Resources"] = channel_status[idx - 1]["Resources"]
-            else:
-                row["Resources"] = dict(self.STATUS_RESOURCE.findall(resources))
+            base_split = self.BASE_RE.findall(row.get("Base"))
+            assert base_split, "Failed to split base into name, channel, and arch"
+            row["Base"] = dict(zip(["name", "channel", "architecture"], base_split[0]))
```
addyess, line 280: hmm... hold off on merge, this breaks bundle builds/promotes entirely

jobs/build-charms/charms.py
```diff
@@ -269,18 +270,24 @@ def info(name, **query):
 
     def status(self, charm_entity):
         """Read CLI Table output from charmcraft status and parse."""
-        charm_status = self.charmcraft.status(charm_entity)
+        charm_status = self.charmcraft.status(charm_entity, _out=None)
```
addyess, line 273: this silences the output of `charmcraft status <charm>` since it's obnoxious how big it is and will soon be replaced with a new version of this command where we parse the json format.

jobs/build-charms/charms.py
```diff
@@ -269,18 +270,26 @@ def info(name, **query):
...
-                row["Resources"] = []
-            elif resources == "↑":
-                row["Resources"] = channel_status[idx - 1]["Resources"]
-            else:
-                row["Resources"] = dict(self.STATUS_RESOURCE.findall(resources))
+            base_split = self.BASE_RE.findall(row.get("Base"))
+            if base_split:
+                row["Base"] = dict(
+                    zip(["name", "channel", "architecture"], base_split[0])
+                )
```
addyess, line 282: `ubuntu 16.04 (arm64)` becomes 

```yaml
   name:  ubuntu
   channel: '16.04'
   architecture: arm64
```

jobs/build-charms/charms.py
```diff
@@ -269,18 +270,26 @@ def info(name, **query):
...
+            for prop in ["Resources", "Revision", "Version"]:
+                value = row.get(prop, "")
+                if value == "↑":
+                    row[prop] = channel_status[idx - 1][prop]
+                elif value == "-":
+                    row[prop] = {} if prop == "Resources" else None
+                elif prop == "Resources":
+                    row[prop] = dict(self.STATUS_RESOURCE.findall(value))
+                else:
+                    row[prop] = value
```
addyess, line 292: `charmcraft status` shows a lot more than it used to, so we're gonna make sure all the ↑ and - and such gets converted appropriately

jobs/build-charms/charms.py
```diff
@@ -337,30 +346,40 @@ def promote(self, charm_entity, from_channel, to_channels):
...
             charm_status = [
                 row
                 for row in self.status(charm_entity)
-                if f"{row['Track']}/{row['Channel']}" == from_channel
+                if all(
+                    (
+                        row["Revision"],
+                        f"{row['Track']}/{row['Channel']}" == from_channel,
+                    )
+                )
```
addyess, line 360: How about this?  Is this easier to read?

```suggestion
                if row["Revision"] and f"{row['Track']}/{row['Channel']}" == from_channel
```

jobs/build-charms/charms.py
```diff
@@ -337,30 +346,40 @@ def promote(self, charm_entity, from_channel, to_channels):
...
             ]
 
         calls = set()
         for row in charm_status:
-            revision, resources = row["Revision"], row["Resources"]
+            revision, resources = int(row["Revision"]), row["Resources"]
```
addyess, line 365: sort on integers and not strings!

jobs/build-charms/charms.py
```diff
@@ -337,30 +346,40 @@ def promote(self, charm_entity, from_channel, to_channels):
...
                     *(f"--channel={chan}" for chan in to_channels),
                     *resource_args,
                 )
             )
-        for args in calls:
+
+        # Act on the charm with the highest revision number
+        # This should always be the most recently built charm
+        for _, *args in sorted(calls)[-1:]:
+            # self._echo(" ".join(["charmcraft", "release", *args]))
```
addyess, line 382: this is for local debugging where you don't REALLY want to run the `chamcraft release`

jobs/build-charms/charms.py
```diff
@@ -523,14 +542,37 @@ def save(self):
...
+        track = self.db["build_args"].get("track") or "latest"
+        if from_channel.lower() in RISKS:
+            from_channel = f"{track}/{from_channel.lower()}"
+        assert (
+            from_channel != "unpublished"
+        ), "It's unwise to promote unpublished charms."
+        ch_channels = [
+            f"{track}/{chan.lower()}" if (chan.lower() in RISKS) else chan
+            for chan in to_channels
+        ]
```
addyess, line 556: this helps assume that `candidate` really means "latest/candidate"

jobs/build-charms/charms.py
```diff
@@ -523,14 +542,37 @@ def save(self):
...
+        failed_entities = []
         for charm_map in self.artifacts:
             for charm_name, charm_opts in charm_map.items():
                 if not any(match in self.filter_by_tag for match in charm_opts["tags"]):
                     continue
-                to_channels = self.apply_channel_bounds(charm_name, to_channels)
-                _CharmHub(self).promote(charm_name, from_channel, to_channels)
+                ch_channels = self.apply_channel_bounds(charm_name, ch_channels)
```
addyess, line 562: this makes sure we don't promote a charm which doesn't support a certain channel due to it's min/min range configuration

jobs/build-charms/charms.py
```diff
@@ -523,14 +542,37 @@ def save(self):
...
+                try:
+                    _CharmHub(self).promote(charm_name, from_channel, ch_channels)
+                except Exception:
+                    self.echo(traceback.format_exc())
+                    failed_entities.append(charm_name)
```
addyess, line 567: this lets us promote as many charms as possible without stopping just because one of them fails

jobs/build-charms/charms.py
```diff
@@ -523,14 +542,37 @@ def save(self):
...
+
+        if any(failed_entities):
+            count = len(failed_entities)
+            plural = "s" if count > 1 else ""
+            raise SystemExit(
+                f"Encountered {count} Promote All Failure{plural}:\n\t"
+                + ", ".join(failed_entities)
+            )
```
addyess, line 575: okay, if there was a failure, report that now

stonepreston (approved):

PR #989, opened 2022-08-24
LP#1983990

mateoflorido:
## Changes
- Fix race condition on Pod creation

### DIFF
--- a/jobs/integration/test_service_endpoints.py
+++ b/jobs/integration/test_service_endpoints.py
@@ -2,7 +2,8 @@ import sh
 import yaml
 from .utils import retry_async_with_timeout, juju_run
 
-SVC_PORT = 50000
+APP_PORT = 50000
+SVC_PORT = 80
 
 
 def get_pod_yaml():
@@ -25,7 +26,7 @@ async def is_pod_running():
         # Pod not created correctly
         return False
 
-    if "Running" in phase and str(SVC_PORT) == port:
+    if "Running" in phase and str(APP_PORT) == port:
         return len(pod["items"]) == 1
 
     # Pod has not fully come up yet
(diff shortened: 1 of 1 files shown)


jobs/integration/test_service_endpoints.py
```diff
@@ -18,11 +18,15 @@ async def is_pod_running():
...
+        _ = pod["items"][0]["spec"]["containers"][0]["env"]
     except IndexError:
         # Pod is not created yet
         return False
+    except KeyError:
+        # Updating Pod envvars
+        return False
 
-    if "Running" in phase:
+    if "Running" in phase and len(pod["items"]) == 1:
```
addyess, line 29: maybe it's best to test something like this

```python
SVC_PORT = 50000
    ...
    try:
        phase = ...
        port = pod["items"][0]["spec"]["containers"][0]["env"]["??PORT??"]
    except (IndexError, KeyError):
        # Pod not created correctly
        return False
        
    if "Running" in phase and str(SVC_PORT) == port:
    ...
```

Cynerva: I like the suggested change to check the port, but we need to keep the `len(pod["items"]) == 1` check too. It's needed to prevent a situation where there exists a `pod["items"][1]` that is listening on the wrong port and causing connections to fail.

mateoflorido: Thanks for the comments and suggestions @addyess @Cynerva. I'll include your suggestions and update the PR.

jobs/integration/test_service_endpoints.py
```diff
@@ -2,6 +2,8 @@
 import yaml
 from .utils import retry_async_with_timeout, juju_run
 
+SVC_PORT = 50000
```
addyess, line 5: toss this port number in everywhere it's mentioned in here.  I see some on lines 49 and 59

jobs/integration/test_service_endpoints.py
```diff
@@ -18,12 +20,14 @@ async def is_pod_running():
...
-        # Pod is not created yet
+        port = pod["items"][0]["spec"]["containers"][0]["env"][0]["value"]
+    except (IndexError, KeyError):
+        # Pod not created correctly
         return False
 
-    if "Running" in phase:
-        return True
+    if "Running" in phase and str(SVC_PORT) == port:
+        return len(pod["items"]) == 1
```
addyess, line 29: oh this is nice

addyess (approved):
LGTM!

kwmonroe (approved):

PR #991, opened 2022-08-25
Set up snap enums for 1.25 stable release

Cynerva:
Bump K8S_NEXT_VERSION to stop us getting 1.25 pre-release builds like `1.25.1-rc.0`.

Add 1.25/stable to 1.25 release channels.


addyess (approved):
Yep.  I'm on board with this.  Should open the 1.25.0 channel

PR #999, opened 2022-08-29
make examples and flow more consistent with updates for the 1.25 release

kwmonroe:
I re-organized some sections based on the order of things we need to do in a new stable release.  It's a pretty big diff, though most of that is due to the content moving around.

Suggest to read/review this as a new document rather than one with 100 diffs.


mateoflorido (approved):
LGTM! 🫶

PR #1000, opened 2022-08-29
Remove the `ConfigObj` dependency

addyess:
Thanks to @Courtney-E-Miller

It was pointed out that this dependency is no longer maintained.  Swapping to parsing the `ConfigParser` built-in rather than an external dependency


mateoflorido (approved):
LGTM! 🥳

PR #1002, opened 2022-08-30
Adds gatekeeper controller manager to support matrix

stonepreston:
Adds support for building/releasing the gatekeeper-controller-manager charm. 

Since this charm is similar to metallb in that the source repo houses multiple charms, the metallb yaml was used as a template. The yaml includes the subdir field to facilitate building the charm from the correct subdirectory.


addyess (approved):
+1 LGTM

PR #1004, opened 2022-08-31
Add gatekeeper tests

nsklikas:
Adds tests for the gatekeeper charms.

The `gatekeeper-audit` charm requires storage, this means that I had to define a `storageClass` and a juju `storage-pool`. I used the `vsphere` provisioner assuming that the jobs will run on vsphere, is this correct?

The tests are not only testing the correctness of each charm on its own, but also the integration of one charm with the other (ie. deploy first one charm and then the other, remove one charm, deploy -> apply policy -> deploy the other, etc)

### DIFF
--- a/jobs/integration/test_opa.py
+++ b/jobs/integration/test_opa.py
@@ -10,7 +10,6 @@ from .logger import log
 from .utils import (
     JujuRunError,
     kubectl,
-    scp_to,
     wait_for_application_status,
     kubectl_apply,
     kubectl_delete,
@@ -291,7 +290,7 @@ class OPATestBase:
 
         try:
             await block_until_with_coroutine(check_app_missing, timeout=120)
-        except asyncio.TimeoutError as e:
+        except asyncio.TimeoutError:
             raise AssertionError(f"Application {app_name} was not removed")
 
 
@@ -299,7 +298,6 @@ class OPATestBase:
 class TestOPAWebhook(OPATestBase):
     async def test_opa_webhook_ready(self, opa_controller_manager):
         await self.wait_for_units(opa_controller_manager)
-        unit = opa_controller_manager.units[0]
 
         log("Waiting for gatekeeper charm to be ready")
         await wait_for_application_status(
--- a/jobs/integration/test_opa.py
+++ b/jobs/integration/test_opa.py
@@ -317,7 +317,6 @@ class TestOPA:
                 "gatekeeper-controller-manager",
             )
 
-
     async def test_opa_audit(self, storage_pool):
         log("Deploying the gatekeeper charm")
         audit = await self.k8s_model.deploy(
--- a/jobs/integration/test_opa.py
+++ b/jobs/integration/test_opa.py
@@ -3,7 +3,6 @@ import json
 import random
 from pathlib import Path
 
-from juju.utils import block_until_with_coroutine
 import pytest
 
 from .logger import log
@@ -13,11 +12,13 @@ from .utils import (
     wait_for_application_status,
     kubectl_apply,
     kubectl_delete,
+    juju_run_action,
 )
 
 templates = Path(__file__).parent / "templates"
 
 
+@pytest.mark.clouds(["vsphere"])
 class OPATestBase:
     @pytest.fixture(autouse=True)
     def setup(self, model, k8s_model, tools):
@@ -81,24 +82,16 @@ class OPATestBase:
         cmd = f"create ns {name}"
         # Creating the namespace should raise
         log("Creating namespace without any labels")
-        try:
+        with pytest.raises(JujuRunError, match="Forbidden|InternalError") as exc_info:
             await kubectl(self.model, cmd)
-        except JujuRunError as e:
-            err = e.output
-            if err.startswith(
-                "Error from server (InternalError): Internal error occurred"
-            ):
-                # Try again as it might be transient/initialization related
-                await asyncio.sleep(10)
-                try:
-                    await kubectl(self.model, cmd)
-                except JujuRunError as e:
-                    err = e.output
-                else:
-                    pytest.fail("Creating the namespace should fail, but it didn't")
-        else:
-            pytest.fail("Creating the namespace should fail, but it didn't")
-        assert err.startswith(
+
+        if "InternalError" in str(exc_info.value):
+            await asyncio.sleep(10)
+            # Try again as it might be transient/initialization related
+            with pytest.raises(JujuRunError, match="Forbidden") as exc_info:
+                await kubectl(self.model, cmd)
+
+        assert exc_info.value.output.startswith(
             "Error from server (Forbidden): admission webhook "
             '"validation.gatekeeper.sh" denied the request:'
         )
@@ -173,32 +166,14 @@ class OPATestBase:
 
     @pytest.fixture(scope="class")
     async def update_status_interval(self, tools, k8s_model):
-        interval = await tools.run(
-            "juju",
-            "model-config",
-            "-m",
-            tools.k8s_connection,
-            "update-status-hook-interval",
-        )
-        interval = interval[0].strip()
+        model_config = await k8s_model.get_config()
+        interval = model_config.get("update-status-hook-interval").value
         try:
             log("Decrease update status hook interval")
-            await tools.run(
-                "juju",
-                "model-config",
-                "-m",
-                tools.k8s_connection,
-                "update-status-hook-interval=5s",
-            )
+            await k8s_model.set_config({"update-status-hook-interval": "5s"})
             yield
         finally:
-            await tools.run(
-                "juju",
-                "model-config",
-                "-m",
-                tools.k8s_connection,
-                f"update-status-hook-interval={interval}",
-            )
+            await k8s_model.set_config({"update-status-hook-interval": interval})
 
     @pytest.fixture(scope="class")
     async def opa_controller_manager(self, model, k8s_model, tools):
@@ -212,15 +187,8 @@ class OPATestBase:
             yield webhook
         finally:
             log("Deleting the gatekeeper charm")
-            await tools.run(
-                "juju",
-                "remove-application",
-                "-m",
-                tools.k8s_connection,
-                "--force",
-                "gatekeeper-controller-manager",
-            )
-            await self.wait_for_app_removed("gatekeeper-controller-manager", k8s_model)
+            await k8s_model.remove_application(webhook.name)
+            await tools.juju_wait(m=tools.k8s_connection, timeout_secs=120)
 
     @pytest.fixture(scope="class")
     async def opa_audit(self, model, k8s_model, tools, storage_pool):
@@ -245,14 +213,13 @@ class OPATestBase:
                 "--destroy-storage",
                 "gatekeeper-audit",
             )
-            await self.wait_for_app_removed("gatekeeper-audit", k8s_model)
+            await tools.juju_wait(m=tools.k8s_connection, timeout_secs=120)
 
     async def _validate_audit_actions(self, unit):
         log("Running list-violations action")
         total_violations = 0
         while total_violations == 0:
-            action = await unit.run_action("list-violations")
-            await action.wait()
+            action = await juju_run_action(unit, "list-violations")
             assert action.status == "completed"
             assert "constraint-violations" in action.results
             constraint_violations = json.loads(action.results["constraint-violations"])
@@ -261,14 +228,14 @@ class OPATestBase:
             await asyncio.sleep(5)
 
         log("Running get-violation action")
-        action = await unit.run_action(
+        action = await juju_run_action(
+            unit,
             "get-violation",
             **{
                 "constraint-template": "K8sRequiredLabels",
                 "constraint": "ns-must-have-gk",
             },
         )
-        await action.wait()
         assert len(json.loads(action.results["violations"])) > 0
 
     async def wait_for_units(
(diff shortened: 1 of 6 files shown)


mateoflorido (changes requested):
Thanks @nsklikas for the PR! I have left a few comments/questions/suggestions:

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,419 @@
...
+                    assert status != "error"
+                    await asyncio.sleep(5)
+            finally:
+                await self.tools.run(
+                    "juju",
+                    "model-config",
+                    "-m",
+                    self.tools.k8s_connection,
+                    f"update-status-hook-interval={interval}",
+                )
```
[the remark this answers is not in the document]
nsklikas: This is used to reset the `update-status-hook-interval`, we set its value in https://github.com/charmed-kubernetes/jenkins/pull/1004/files#diff-1b759efbcf98e4a258b9b9ed2a9e7ff23d25298e30f071854a85803332791467R260-R266.

Is it ok if we leave the `update-status-hook-interval` value altered?

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,419 @@
...
+
+            try:
+                await self.deploy_example_policy()
+                log("Test that the policy is enforced")
+                await self.validate_create_ns_fail()
+                await self.validate_create_ns_with_label()
+                await self.destroy_example_policy()
+                log("Validate that the policy is no longer enforced")
+                await self.validate_create_ns()
+                self._validate_audit_actions(unit)
```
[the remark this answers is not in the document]
nsklikas: You are right, good catch

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,419 @@
...
+            await self.destroy_example_policy()
+        finally:
+            log("Deleting the gatekeeper charm")
+            await self.tools.run(
+                "juju",
+                "remove-application",
+                "-m",
+                self.tools.k8s_connection,
+                "gatekeeper-controller-manager",
+            )
```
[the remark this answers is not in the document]
nsklikas: I delete it because the fixture deletes the applications after the tests have run, but I want the `gatekeeper-controller-manager` gone for the next test to test the audit charm on its own.

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,419 @@
...
+        await action.wait()
+        assert len(json.loads(action.results["violations"])) > 0
+
+    async def test_opa_webhook(self, storage_pool):
+        log("Deploying the gatekeeper charm")
+        webhook = await self.k8s_model.deploy(
+            "gatekeeper-controller-manager",
+            channel=self.tools.charm_channel,
+            trust=True,
+        )
```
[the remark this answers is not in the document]
nsklikas: The fixture would be used only for one test, I don't think it would offer much.

mateoflorido (changes requested):

jobs/integration/test_opa.py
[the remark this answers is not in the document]
nsklikas: The `k8s_model` fixture is used in other tests as well, to change the fixture's scope we would need to copy it, I think it's easier to simply clean up the model.

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,408 @@
...
+            self.tools.controller_name,
+            self.tools.connection,
+        )
+        log("Delete example policy")
+        await kubectl(self.model, f"delete -f {remote_path}")
+
+    @pytest.fixture(scope="module")
+    async def storage_pool(self, model, k8s_model, tools):
+        # This assumes that we are running on vsphere, should it be
+        # extended to run on other clouds as well?
```
mateoflorido, line 134: @kwmonroe As far as I know, we test on both vSphere and AWS. Is this something that needs to be addressed here?

mateoflorido (approved):
LGTM, I just have one question for @kwmonroe.
Test results: [vSphere](https://pastebin.ubuntu.com/p/ThZZ5DZCmG/)

jobs/integration/test_opa.py
nsklikas: As discussed, I added the `vsphere` mark, should be fine now

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,408 @@
...
+            self.control_plane_unit,
+            remote_path,
+            self.tools.controller_name,
+            self.tools.connection,
+        )
+
+        # Create the namespace
+        cmd = f"apply -f {remote_path}"
+        log("Creating namespace")
+        await kubectl(self.model, cmd)
```
addyess, line 36: maybe there should be a util method called `kubectl_apply(document: Path)` which can do this for you.  Seems this is repeated quite a bit.

addyess: and perhaps its analog is `kubectl_delete(document: Path)`

jobs/integration/test_opa.py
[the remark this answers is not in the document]
addyess: fixtures do aid in guaranteed teardown which is a bit nice than

```python
   try:
     # setup stuff
     # run test 
   finally:
     # teardown
```

i think @mateoflorido is on to something when he's advocating for a fixture here.  You can get a test ERROR because of a setup/teardown failure and test FAILED when the validation fails.

i'm not a fan of the style of nested try blocks, but it seems like it will work, but this test method is really way too big.  It's not testing one thing -- but everything. 


why not `test_opa_charm_ready`, `test_validate_create_ns_no_policy`, `test_validate_create_ns_policy_enforcement` ... etc?

then that fixture becomes a `module` based fixture that runs one, is used by each small test, and then cleans up when all the tests in the module are run.

addyess (changes requested):
Overall, my only suggestions here are to break the 2 big tests down into smaller tests, and maybe do consider fixtures for setup/teardown.  if you don't like module based fixtures -- you could do class based fixtures 

your two current tests could be broken into two classes
those two classes could have multiple tests
and you could have class scoped fixtures to control their setup/teardown.

jobs/integration/utils.py
```diff
@@ -595,6 +610,30 @@ async def kubectl(model, cmd, check=True):
     )
 
 
+async def _kubectl_doc(document, control_plane_unit, tools, model, action):
```
addyess, line 613: ❤️

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+        finally:
+            log("Deleting the gatekeeper charm")
+            await tools.run(
+                "juju",
+                "remove-application",
+                "-m",
+                tools.k8s_connection,
+                "--force",
+                "gatekeeper-controller-manager",
+            )
```
addyess, line 222: this should work too

```suggestion
            await k8s_model.remove_application(webhook.name)
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+            log("Deleting the audit charm")
+            await tools.run(
+                "juju",
+                "remove-application",
+                "-m",
+                tools.k8s_connection,
+                "--force",
+                "--destroy-storage",
+                "gatekeeper-audit",
+            )
```
addyess, line 247:
```suggestion
            await k8s_model.remove_application(audit.name)
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+            await self.wait_for_app_removed("gatekeeper-audit", k8s_model)
+
+    async def _validate_audit_actions(self, unit):
+        log("Running list-violations action")
+        total_violations = 0
+        while total_violations == 0:
+            action = await unit.run_action("list-violations")
```
addyess, line 254: there are changes coming between juju 2.9.11 and 3.0.1  that mean we should use `juju_run_action` rather than `unit.run_action`.  I'll save you the hassle but python-lib-juju is making a breaking change to their API so i'm trying to shove all the action running logic into there. 

There's plenty of examples of how to use `juju_run_action(...)` in validation.py

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+        try:
+            await block_until_with_coroutine(check_app_missing, timeout=120)
+        except asyncio.TimeoutError:
+            raise AssertionError(f"Application {app_name} was not removed")
+
+
+@pytest.mark.clouds(["vsphere"])
+class TestOPAWebhook(OPATestBase):
+    async def test_opa_webhook_ready(self, opa_controller_manager):
+        await self.wait_for_units(opa_controller_manager)
```
addyess, line 300: if you don't want to use `juju_wait`, there's also 

```python
   # waits for all units in the model to be idle/active
   await self.k8s_model.wait_for_idle(wait_for_active=True, timeout=120)
   
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+        await kubectl(self.model, cmd)
+
+    async def validate_create_ns_fail(self, name=None):
+        if not name:
+            name = f"test-ns-{random.randint(1, 99999)}"
+
+        cmd = f"create ns {name}"
+        # Creating the namespace should raise
+        log("Creating namespace without any labels")
+        try:
```
addyess, line 84: [pytest](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-raises) tells me that testing for expected exceptions should look like this:

```python
with pytest.raises(JujuRunError) as ie:
    await kubectl(self.model, cmd)
```

now you've got an interesting wrinkle in here that says to retry for transient errors.  

So how about this, it's still a bit repetitive but it makes use of the `match` method in pytest.raises

```python
with pytest.raises(JujuRunError, match="Forbidden|InternalError") as exc_info:
    await kubectl(self.model, cmd)
    
if "InternalError" in str(exc_info.value):
    await asyncio.sleep(10)
    with pytest.raises(JujuRunError, match="Forbidden") as exc_info:
        await kubectl(self.model, cmd)
        
assert exc_info.output.startswith("Error from server (Forbidden) ...")
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+    @pytest.fixture(scope="class")
+    async def update_status_interval(self, tools, k8s_model):
+        interval = await tools.run(
+            "juju",
+            "model-config",
+            "-m",
+            tools.k8s_connection,
+            "update-status-hook-interval",
+        )
+        interval = interval[0].strip()
```
addyess, line 183:
```suggestion
        model_config = await k8s_model.get_config()
        interval = model_config.get("update-status-hook-interval").value
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+        try:
+            log("Decrease update status hook interval")
+            await tools.run(
+                "juju",
+                "model-config",
+                "-m",
+                tools.k8s_connection,
+                "update-status-hook-interval=5s",
+            )
```
addyess, line 192:
```suggestion
            await k8s_model.set_config({"update-status-hook-interval": "5s"})
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+            yield
+        finally:
+            await tools.run(
+                "juju",
+                "model-config",
+                "-m",
+                tools.k8s_connection,
+                f"update-status-hook-interval={interval}",
+            )
```
addyess, line 201:
```suggestion
            await k8s_model.set_config({"update-status-hook-interval": interval})
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+
+    async def test_policy_enforcement(self, opa_controller_manager):
+        log("Creating policy and constraint crds")
+        await self.deploy_example_policy()
+        log("Test that the policy is enforced")
+        try:
+            await self.validate_create_ns_fail()
+            await self.validate_create_ns_with_label()
+        finally:
+            await self.destroy_example_policy()
```
addyess, line 328: this test doesn't assert anything?  Are we relying on the fact that nothing is raised in these `validate` methods?  As long as they have asserts i guess that's okay

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+            await self.tools.run(
+                "juju",
+                "remove-application",
+                "-m",
+                self.tools.k8s_connection,
+                "--force",
+                "--destroy-storage",
+                "gatekeeper-audit",
+            )
+            await self.wait_for_app_removed("gatekeeper-audit")
```
addyess, line 365:
```suggestion
            await self.k8s_model.remove_application(audit.name)
            await self.tools.juju_wait(...)
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,439 @@
...
+
+    async def test_manager_reconcile(
+        self, opa_controller_manager, update_status_interval
+    ):
+        await self.wait_for_units(
+            opa_controller_manager, error_status=("error",), expected_status="blocked"
+        )
```
addyess, line 372: ahh -- looking for blocked units.  okay now i see the value of `wait_for_units`.

jobs/integration/test_opa.py
nsklikas: This does not remove storage, I thought it's simpler to run the juju command rather than having to fetch the used storages and free them before to removing the storage-pool

jobs/integration/test_opa.py
nsklikas: the `validate_*` assert that the correct errors are returned (or not returned)

jobs/integration/test_opa.py
addyess: ok, how about 

```python
k8s_model.remove_application(audit.name, destroy_storage=True, force=True)
```

addyess: https://github.com/juju/python-libjuju/blob/master/juju/model.py#L1008

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,392 @@
...
+            log("Deleting the audit charm")
+            await self.tools.run(
+                "juju",
+                "remove-application",
+                "-m",
+                self.tools.k8s_connection,
+                "--force",
+                "--destroy-storage",
+                "gatekeeper-audit",
+            )
```
addyess, line 320: Could do this

```suggestion
            await self.k8s_model.remove_application(audit.name, force=True, destroy_storage=True, block_until_done=True)
```

jobs/integration/test_opa.py
```diff
@@ -0,0 +1,392 @@
...
+            yield storage_pool_name
+        finally:
+            log("Removing storage pool")
+            await tools.run(
+                "juju",
+                "remove-storage-pool",
+                "-m",
+                tools.k8s_connection,
+                storage_pool_name,
+            )
```
addyess, line 165: and this:
```suggestion
            await k8s_model.remove_storage_pool(storage_pool_name)
```

addyess (approved):
Couple of small nits but this looks pretty nice!

jobs/integration/test_opa.py
nsklikas: It seems like these options were added in a later version(https://github.com/juju/python-libjuju/blob/2.9.11/juju/model.py#L878), maybe leave it like this with a comment to change it when we update `python-libjuju`'s version?

addyess: AH!!  you're 100% right! Same comment for storage things below

PR #1005, opened 2022-08-31
ensure oci-images are sunk with 1.25

kwmonroe:
We were missing a `1.25-upstream` line in our `container-images.txt` file (in the bundle repo), which led to an incomplete container listing for the 1.25 release.  This is because we didn't sync upstream oci images for 1.25.  Let's do that.


stonepreston (approved):
LGTM

PR #1013, opened 2022-09-08
test_status_message handle casses with gpu not enabled

asbalderson:
There is a second message possible from a running
kubernetes-worker, "Kubernetes worker running (without gpu support)"

Use `in` to check that "Kuberentes worker running" is in the message to handle both cases.


asbalderson:
I had originally opted to use check a list of stats messages using `any` but went with this solution because when the assertion fails, it reports the current message, instead of a list of "False" messages.

Basdbruijne (approved):
+1

addyess (approved):
Nice. Thanks for the contribution

PR #1015, opened 2022-09-09
update template language; add a jjb checkbox

kwmonroe:
Update our jenkins PR template with accurate locations and add a checkbox if submitters know that a PR will need a `jjb` run.


addyess (approved):
nice!  like the jjb add

PR #1016, opened 2022-09-09
Allow HTTPError to be raised when branch or tag doesn't exist in github

addyess:


kwmonroe (approved):
LGTM, but did you mean to include the k8s-storage min version changes in this PR?

PR #1017, opened 2022-09-09
use the layer/charm options for minimum channel support in the release jobs

addyess:
Moves the `ChannelRange` object and `Release` objects to `cilib.version` in order to spread its usage between build_charms and the other release jobs.  

any layer or charm can specify an option like this:

```yaml
    channel-range:
      min: '1.25'
      max: '1.30'
```

which bounds its channel range.  Both min and max are optional. 

Previously, these were only used by charm build job, but it's necessary to also be able to use these in other jobs (such as sync)

### DIFF
--- a/jobs/sync-upstream/sync.py
+++ b/jobs/sync-upstream/sync.py
@@ -187,7 +187,7 @@ def _rename_branch(
                 log.info(f"Skipping  :: {layer_name:^40} :: {to_name} already exists")
                 continue
 
-            log.info(f"Renaming :: {layer_name:^40} :: from: {from_name} to:{to_name}")
+            log.info(f"Renaming  :: {layer_name:^40} :: from: {from_name} to:{to_name}")
 
             try:
                 repo.rename_branch(from_name, to_name)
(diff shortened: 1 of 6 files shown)


jobs/sync-upstream/sync.py
```diff
@@ -156,22 +172,23 @@ def _rename_branch(
...
-                log.info(f"Skipping  :: {layer_name:^35} :: {from_name} doesn't exist")
+                log.info(f"Skipping  :: {layer_name:^40} :: {from_name} doesn't exist")
                 continue
 
             if to_name in repo.branches:
-                log.info(f"Skipping  :: {layer_name:^35} :: {to_name} already exists")
+                log.info(f"Skipping  :: {layer_name:^40} :: {to_name} already exists")
                 continue
 
+            log.info(f"Renaming :: {layer_name:^40} :: from: {from_name} to:{to_name}")
```
stonepreston, line 190:
```suggestion
            log.info(f"Renaming  :: {layer_name:^40} :: from: {from_name} to:{to_name}")
```
Small formatting fix to get all the :: lined up

stonepreston (approved):
LGTM, suggested change to align the rename line with the others

jobs/sync-upstream/sync.py
addyess: you have some sort of eagle eyes 🦅

PR #1018, opened 2022-09-09
set 1.25 as the default stable

kwmonroe:
- set current stable version to 1.25
- drop <= 1.22 from our matrix where appropriate (most tests only need to run against n-2)
- bionic is covered by `validate-ck-stable`; drop it for other jobs to reduce run time
- snap tracks [are open](https://forum.snapcraft.io/t/kubernetes-1-26-snap-tracks/31491) for 1.26; start building them

---

- [x] Needs `jjb` after merge
  - `[build-snaps|release|validate].yaml`


addyess (approved):
Looks good to me!

PR #1031, opened 2022-09-16
Tag charm repos with charmhub revision after charm push

addyess:


jobs/build-charms/charms.py
```diff
@@ -758,6 +772,11 @@ def push(self):
             f"Pushing {self.type}({self.name}) from {self.dst_path} to {self.entity}"
         )
         self.new_entity = _CharmHub(self).upload(self.dst_path)
+        if repo := self.repository:
+            tag = f"{self.name}-{self.new_entity}"
+            message = f"Built by job: {self.build.build_tag}"
+            self.echo(f"Tagging {self.type}({self.name}) with {tag}")
+            repo.tag_commit(self.commit(), tag=tag, message=message)
```
stonepreston, line 779: If we end up doing more stuff with these build tags in the future, we will need to remember that certain repos (the ones with multiple charms in a repo like metallb and opa-gatekeeper) will end up with multiple of these build tags on the same commit. I dont think there are any issues with that happening, but something we should keep in mind and maybe call out with a comment in case we end up writing code that consumes the build tags in the future, and note that it wont always be a 1 to 1 relationship with commits and build tags.

stonepreston (approved):
LGTM, but left some thoughts on adding a comment clarifying the relationship between a commit and the possibility of multiple tags existing for it.

jobs/build-charms/charms.py
addyess: hmm... actually it could be worse than that.  You can't tag a repo twice with the same tag.  This could result in build failures simply because it can't tag the repo.  Lemme see if there's an adjustment to be made.  

Here's what i'm thinking:
* Tag if-and-only-if the tag doesn't yet exist in the repo
* If it DOES exist already, compare the tag hash to the current commit hash
* Error if the tag exists on another commit (Somehow charmhub assigned a duplicate revision? )

- Do we care if a single commit has 20 or 30 of the same tag?  ie -- the repo didn't change but maybe a reactive layer changed or a build was forced

stonepreston: yeah I thought that could be an issue, but thought the tags would be unique since the "duplicate" tag would have a different charm name

addyess: Well, clearly this needs more thought.

stonepreston: as to the multiple tags on a single commit, I think that would be desirable, as its a tie between jenkins jobs and specific commits. IE if a build goes out and builds a new charm due to a layer change but NOT due to a charm code change, id still like to be able to go back and see that this job was using this commit as its source.

addyess:
@stonepreston after some thought:
* this still supports multiple tags on a single commit
* this now supports the same tag on the same commit (for charms/bundles which supports any retagging that might happen)
* this errors out on an entity which wants to apply over an existing tag
* this now uses lightweight tags rather than annotated tags
* this now uses tag pagination for when the tags grow to large numbers of tags

stonepreston (approved):

PR #1033, opened 2022-09-19
address issues with not parsing stdout from `charm pull-source ...`

addyess:
`charm pull-source` was removed from the charm tool inadvertently and reintroduced 
* https://github.com/juju/charm-tools/pull/626 -- removed
* https://github.com/juju/charm-tools/pull/628 -- readded

* These changes allow the build ci to actually show the stdout when a `charm` command fails as it did in CI so we can trace the point of failure

- [ ] needs refresh with jjb


stonepreston (commented):
This LGTM, but can you provide some background as to what the issue was with the parsing?

stonepreston (approved):

PR #1036, opened 2022-09-21
upgrade protobuf from 3.15.6 -> 3.20.1

addyess:
I found that upgrading this package removed so many warnings from the validation runs.

---

- [ ] Needs `jjb` after merge


addyess:
🤠

PR #1040, opened 2022-09-26
rename integrator validate model and update overlay config

addyess:
Fix issues in validate-ck-integrator job on its first jenkins runs

---

- [x] Needs `jjb` after merge: validate.yaml


kwmonroe (approved):
LGTM; looking forward to seeing the parallel jobs in action.

PR #1041, opened 2022-09-27
pin pack charm-tools until all reactive charms can build with charmcraft

addyess:
Pins charm-tools to 2.x branch until the builds can be migrated to using charmcraft. 

- [ ] Needs `jjb` after merge


stonepreston (approved):
LGTM once the snap lands in the new channel

PR #1044, opened 2022-09-28
Added creation of aws resources for eksd testing

berkayoz:
---

- [ ] Needs `jjb` after merge

Added creation of aws resources for eksd testing for MicroK8s releases. The resources are created accordingly and cleaned up after the tests completion.


neoaggelos (commented):
As a comment, I would _very much_ prefer to use a CloudFormation template for this instead. That way, creation commands are idempotent, and it is much easier to track the resources and re-use them in other places (e.g. setting up a testing environment manually).

jobs/infra/playbook-jenkins.yml
```diff
@@ -164,11 +163,22 @@
...
       loop:
+        - "aws-cli"
         - "microk8s"
         - "bundletester"
         - "surl"
       tags:
         - jenkins
+    - name: install awscli
+      shell: |
+        curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
kwmonroe, line 174: this will run on all jenkins workers (amd, arm, s390x).  if you *really* only want amd64 installed, you should wrap this shell command in a "if arch == amd64".

also consider the [more recent amz-aws-cli snap](https://snapcraft.io/amz-aws-cli) if you don't want to go through this slightly more complicated shell script.

jobs/release-microk8s/release-microk8s.groovy
```diff
@@ -11,13 +11,14 @@ def destroy_controller(controller) {
 
 pipeline {
     agent {
-        label "runner-${params.ARCH}"
+        label "runner-amd64-2"
```
kwmonroe, line 14: why limit this to `amd64-2`?  won't the [arm64 job](https://jenkins.canonical.com/k8s/job/release-microk8s-arch-arm64/) have problems if run on amd64?

If you *do* always want this on amd64, i'd pick one of the [`ps5-xx` workers](https://jenkins.canonical.com/k8s/computer/).  They're way faster.

kwmonroe (changes requested):
Couple questions / concerns in-line below.

jobs/release-microk8s/release-microk8s.groovy
berkayoz: Good catch, thank you. This was a testing artifact, fixed.

jobs/infra/playbook-jenkins.yml
berkayoz: I have questions regarding this. I would much prefer using the more recent snap, but maintenance is an unknown. And I believe this snap is also just for `amd64`. 

We would need these calls in arm as well, could we make a snap and maintain it ourselves? But I believe there are only amd64 and arm installers from aws for the cli v2. Maybe using the `awscli` [apt package](https://packages.ubuntu.com/jammy/awscli) is better, since it supports all architectures. Only downside we will be following an older version (1.22 something currently).

@ktsakalozos I would like to hear your thoughts on this as well. I think the apt package would work for us.

ktsakalozos: @berkayoz lets go with the apt package. Getting the snap will limit us to amd64 and always getting the latest from the .zip may introduce some flakiness.

ktsakalozos:
All review comments seem addressed. Merging this.

PR #1047, opened 2022-09-29
nagios validation times out when nagios alerts don't match expected alert states

addyess:


jobs/integration/utils.py
```diff
@@ -183,13 +183,14 @@ async def scp_to(local_path, unit, remote_path, controller_name, connection_name
 async def retry_async_with_timeout(
     func,
     args,
+    kwds=None,
```
addyess, line 186: the async method in `func` will be called with `*args` -- so let's add `kwds` here so that async can have keyword-arguments as well.

jobs/integration/utils.py
```diff
@@ -199,12 +200,13 @@ async def retry_async_with_timeout(
 
     """
     deadline = time.time() + timeout_insec
+    results = None
```
addyess, line 203: the async `func` can return anything that is truthy.  objects can be truthy if they implement the `__bool__` magic method.  Knowing this, the `timeout_msg` could have a format marker to print the timed out object. 

consider this:

```python

class Truthy:
    def __bool__(self):
        return False

    def __repr__(self):
        return "Truthy()"

async def my_async():
    return Truthy()


...
    await retry_async_with_timeout(
        my_async,
        tuple(),
        timeout_msg="I will timeout because my_async returns {}"
    )
```


This will eventually fail with the expected TimeoutError
but that Exception will have some test describing the timeout exception
`I will timeout because my_async returns Truthy()`

jobs/integration/validation.py
```diff
@@ -2205,32 +2206,82 @@ async def cleanup():
...
+
+    def __str__(self) -> str:
+        return ", ".join(
+            f"{severity}/{link.string}"
+            for severity, alerts in self.alerts.items()
+            for alert in alerts
+            for link in alert.find_all("a", recursive=False)
+        )
+
+
```
addyess, line 2233: this is a `truthy` object that is true when there are some alerts in the mapping and all the matchers are true
if cast to a string it will list all the alerts it has by severity

there's also an accessor method here to `get` the alerts by severity

jobs/integration/validation.py
```diff
@@ -2205,32 +2206,82 @@ async def cleanup():
...
             "ok": "statusOK",
         }
         url_data = self.open(self.url)
         soup = bs(url_data.read(), "html.parser")
-        return {
+        alerts = {
             severity: soup.find_all("td", class_=classes[severity.lower()])
             for severity in severities
         }
+        return NagiosAlerts(alerts, severities)
```
addyess, line 2251: this becomes an async method so it can be used by `retry_async_with_timeout`  It now also returns a `NagiosAlerts` object which can be evaluated for truthiness.

jobs/integration/validation.py
```diff
@@ -2205,32 +2206,82 @@ async def cleanup():
...
+                    for app in apps:
+                        if app in link.string:
+                            found[app].append(link.string)
+            if all(app_crits for app_crits in found.values()):
+                log(f"Found critical errors in {', '.join(apps)}:")
+                for app, app_crits in found.items():
+                    for crit in app_crits:
+                        log(f"{app} - {crit}")
+                return True
+        return False
```
addyess, line 2270: Adding this method that looks for `critical` nagios alerts for a given set of apps.

It only returns `True` when every app given has at least one critical error.   Its just the generic version of what the test was doing in a `while True` loop [down here](https://github.com/charmed-kubernetes/jenkins/pull/1047/files#diff-a3e9600e4339acd8390a865b32d04515eba72f1929856af1e778e36d7e550fb6L2309-L2336)

jobs/integration/validation.py
```diff
@@ -2254,15 +2305,27 @@ async def nagios(model, tools):
...
         await app.expose()
     if "nrpe" not in model.applications:
+        excludes = [
+            "/snap/",
+            "/sys/fs/cgroup",
+            "/run/containerd",
+            "/var/lib/docker",
+            "/run/credentials",
+            "/run/systemd/incoming",
+        ]
```
addyess, line 2315: this overrides the defaut `nrpe` config `ro_filesystem_excludes`. 

it comes with `/snap/,/sys/fs/cgroup,/run/containerd,/var/lib/docker` by default
in jammy, the test wouldn't setup as stable until we ignored the other read-only filesystems that nagios could see:
`/run/credentials` and `/run/systemd/incoming`.  

TBH -- i found `/run/credentials` mounted in read-only on a jammy file system.  I never found `/run/systemd/incoming` in `/proc/mounts` so i have no clue why nagios is seeing it.  But in eithercase, i'm ok if that filesystem is readonly during this tests -- so lets exclude it!

jobs/integration/validation.py
```diff
@@ -2254,15 +2305,27 @@ async def nagios(model, tools):
...
+            ),
             num_units=0,
             channel="stable",
         )
-    for each in model.applications:
-        if each not in ["nrpe"]:
+    for each, related in model.applications.items():
+        if each not in ["nrpe"] and not any(
+            "nrpe:" in str(rel) for rel in related.relations
+        ):
```
addyess, line 2328: if NRPE is already related to the application, skip `re-relating` it to the application.  I found this was necessary while developing the test.  I didn't want to have to keep redeploying nagios and nrpe.  But this was necessary to continue the setup if nrpe was already related to the app

jobs/integration/validation.py
```diff
@@ -2283,7 +2346,10 @@ async def nagios(model, tools):
     # 3) wait for nagios to settle
     log("waiting for nagios to settle")
     nagios_api = NagiosApi(opener.open, status_url)
-    await nagios_api.wait_for_alerts(critical=lambda c: not c, pending=lambda p: not p)
+    await nagios_api.wait_for_settle(
+        stage="after deployment",
+        timeout_insec=60 * 15,
+    )
```
addyess, line 2352: you've got 15 min to have no criticals and no pending alerts after deployment - GO!

jobs/integration/validation.py
```diff
@@ -2306,68 +2372,61 @@ async def test_nagios(model, nagios):
...
+        await retry_async_with_timeout(
+            nagios.critical_alerts_by_app,
+            (
+                "kubernetes-control-plane",
+                "kubernetes-worker",
+            ),
+            timeout_insec=10 * 60,
+            timeout_msg="Failed to find critical errors in control-plane and worker after 10m.",
+            retry_interval_insec=30,
+        )
```
addyess, line 2394: this gets run in a `try...finally`.  If the tests FAILS (due to a timeout maybe) we still want to restore the broken control-plane. 

next notice the test is effectively run in the `critical_alerts_by_app` method.  keep looking for critical alerts in the control plane and worker units or until the 10 minutes is up.  A timeout constitutes a test failure.  But don't leave the control plane broken after the failed test.

jobs/integration/validation.py
```diff
@@ -2306,68 +2372,61 @@ async def test_nagios(model, nagios):
...
 
-    # 7) break worker
-    log("Breaking workers")
-    workers = masters = model.applications["kubernetes-worker"]
-    await workers.set_config({"kubelet-extra-args": "broken=true"})
+    # wait for complaints to go away
+    await nagios.wait_for_settle(
+        stage="restoring API Server",
+        timeout_insec=10 * 60,
+    )
```
addyess, line 2404: if the above test was successful (ie -- found some critical alerts), let's let the API server heal and let nagios settle again.  This is ANOTHER test.  Do the alerts clear in 10 minutes after restoring the api server configs?

jobs/integration/validation.py
```diff
@@ -2306,68 +2372,61 @@ async def test_nagios(model, nagios):
...
+        await retry_async_with_timeout(
+            nagios.critical_alerts_by_app,
+            ("kubernetes-worker",),
+            timeout_insec=10 * 60,
+            timeout_msg="Failed to find critical errors in worker after 10m.",
+            retry_interval_insec=30,
+        )
+    finally:
+        # 9) Fix worker
+        await workers.set_config({"kubelet-extra-args": ""})
```
addyess, line 2423: similar to the first test, we're gonna break the kubelets on all the workers and look for critical failures on only the workers.  It's got 10 minutes to fail -- GO!

jobs/integration/validation.py
```diff
@@ -2306,68 +2372,61 @@ async def test_nagios(model, nagios):
...
 
-    # 9) Fix worker and wait for complaints to go away
-    await workers.set_config({"kubelet-extra-args": ""})
-    await nagios.wait_for_alerts(critical=lambda c: not c)
+    # wait for complaints to go away
+    await nagios.wait_for_settle(
+        stage="restoring worker kubelets",
+        timeout_insec=10 * 60,
+    )
```
addyess, line 2429: ok, the workers should have been restored -- do the alerts clear?

PR #1057, opened 2022-10-11
include azure and gce deployments for integrator charm tests

addyess:
* bootstrap and deploy the integrator tests on azure and gce
* resolve test issues with load balancers on azure and gce
* improve debugging of failed storage tests
* resolve test issues on vsphere

---

- [x] Needs `jjb` after merge

### DIFF
--- a/jobs/validate/integrator-spec
+++ b/jobs/validate/integrator-spec
@@ -51,6 +51,8 @@ EOF
     elif [ "$JUJU_CLOUD" = "aws/us-east-1" ]; then
       # Deploy aws
       cat <<EOF >> overlay.yaml
+  easyrsa:
+    to: null
   aws-integrator:
     charm: aws-integrator
     channel: $JUJU_DEPLOY_CHANNEL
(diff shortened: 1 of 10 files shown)


addyess:
adding azure and google clouds for integrator testing

stonepreston (approved):

jobs/validate/integrator-spec
```diff
@@ -51,6 +51,8 @@ EOF
     elif [ "$JUJU_CLOUD" = "aws/us-east-1" ]; then
       # Deploy aws
       cat <<EOF >> overlay.yaml
+  easyrsa:
+    to: null
```
stonepreston, line 55: what does the to: null do in this case?

addyess: this negates any option set at this field. 

the kubernetes-core bundle sets this on `easyrsa`
```yaml
    easyrsa:
        to:
        - lxd:0
```

this means easyrsa ends up on a lxd on machine 0.  When this occurs, it can't find out information about it's public ip address that it needs to build a cluster certificate.

So -- by removing this config option it will get its own VM in the cloud rather than running in a lxd

stonepreston (approved):

PR #1071, opened 2022-10-21
purge google cloud instances

addyess:
* purge any running instances which were created by k8sci user (across the entire ubuntu-benchmarking project)
* purge any Service Account and User Managed Keys (across the entire ubuntu-benchmarking project)
   * I'm not sure how to filter the service accounts created by the `k8sci` user


kwmonroe (approved):
:rocket:

PR #1072, opened 2022-10-21
upgrade to potential juju==3.0.4 library

addyess:
Addresses running validation tests on juju 3 controllers
[LP#1993533](https://bugs.launchpad.net/juju/+bug/1993533)

### DIFF
--- a/jobs/build-charms/resource-spec.yaml
+++ b/jobs/build-charms/resource-spec.yaml
@@ -31,5 +31,3 @@
   calico-cni-arm64: '{out_path}/calico-cni-arm64.tar.gz'
   calicoctl-image: '{out_path}/calicoctl-image.tar.gz'
   calico-node-image: '{out_path}/calico-node-image.tar.gz'
-'sriov-cni':
-  sriov-cni-image: 'sriov-cni'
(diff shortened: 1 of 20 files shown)


kwmonroe (commented):
I'm a tentative +1 on this.  It all looks sound... for now.  But I know we're facing significant challenges with the `juju 3` snap.  I fear libjuju could change in a way that breaks us before we're ready to move off 2.9.

Perhaps pinning to `3.0.3` would ease those fears. @addyess is there anything in this v3 lib that helps us in a 2.9/stable env?

addyess:
well, this PR does pin at `juju==3.0.3`. 

i tested this change with a juju 2.9.x controller and juju 2.9 snap client on my machine and had passing testes

i also tested this change with a juju 3.x controller and juju 3.x snap client on my machine to watch it fail
-- but i couldn't get the model up to run those tests


I'm okay holding this until a mature juju 3 controller/client are locally functional

kwmonroe (approved):
After prague discussion with juju, i'm comfortable with 3.0.4 in use against both 2.9 and 3.0 controllers.  :rocket:

PR #1088, opened 2022-10-28
Raise test_toggle_metrics timeouts to 10 minutes

Cynerva:
Fixes https://bugs.launchpad.net/bugs/1995139

As noted in the bug, things can get held up if kube-ovn-controller is in CrashLoopBackOff. The maximum backoff is capped at 5 minutes ([source](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy)), so 10 minutes should be enough time to let kube-ovn-controller recover and metrics pods to subsequently settle.

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -1157,7 +1157,6 @@ async def test_audit_default_config(model, tools):
     await reset_audit_config(app, tools)
 
 
-@pytest.mark.flaky(max_runs=5, min_passes=1)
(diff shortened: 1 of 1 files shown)


Cynerva:
I removed `@pytest.mark.flaky` because I suspect the flakiness may have been a symptom of the timeout being too low. I'm curious to see if we have problems now that the timeout has been raised. Let me know if there are any concerns.

PR #1092, opened 2022-11-01
`test_snap_version` handles `latest/edge` charm option

addyess:
if the charm's channel is 'latest/{risk}', test that the running snaps on a unit matches the channel version in snapstore for that 'latest/{risk}' channel


mateoflorido (approved):
LGTM!

PR #1094, opened 2022-11-02
adjust kubernetes-dashboard namespace

addyess:
when the charm channel is set to `latest/{risk}`, the tests should assume the kubernetes-dashboard is no longer in the `kube-system` namespace. 

I could be talked into not testing the dashboard being within that ns as well since we likely won't be testing 1.15 and below

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -386,7 +386,9 @@ async def test_dashboard(model, log_dir, tools):
         k8s_version = (2, 0)
         if "/" in channel:
             version_string = channel.split("/")[0]
-            k8s_version = tuple(int(q) for q in re.findall("[0-9]+", version_string)[:2])
+            k8s_version = tuple(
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -379,18 +379,21 @@ async def query_dashboard(url, config):
...
+        # we should default to the latest dashboard url format
+        k8s_version = (2, 0)
+        if "/" in channel:
+            version_string = channel.split("/")[0]
+            k8s_version = tuple(
+                int(q) for q in re.findall("[0-9]+", version_string)[:2]
+            )
+
+        if k8s_version < (1, 16):
+            dash_ns = "kube-system"
```
addyess, line 394: Or should we just forget about testing k8s versions < `1.16`?

```suggestion
    dash_ns = "kubernetes-dashboard"
```

addyess (commented):
Or we could just do this

mateoflorido (approved):
LGTM :)

PR #1096, opened 2022-11-02
tigera-secure-ee uses overlay to deploy over charmed-kubernetes bundle

addyess:
the `kubernetes-tigera-secure-ee` bundle no longer is published to charmhub, instead use this overlay.


mateoflorido (approved):
LGTM 👍🏻

PR #1100, opened 2022-11-09
upgrade github actions workflows

addyess:
* uses inclusive-workflow callable workflow
* updates `actions/checkout` and `actions/setup-python`
* corrects that each python matrix is tested
* uses setup-python's cache abilities
* runs the tests outside of tox on github actions to make use of the cached python requirements

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,20 +23,11 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
-      - name: Cache pip
-        uses: actions/cache@v3
-        with:
-          # This path is specific to Ubuntu
-          path: ~/.cache/pip
-          # Look to see if there is a cache hit for the corresponding requirements file
-          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
-          restore-keys: |
-            ${{ runner.os }}-pip-
-            ${{ runner.os }}-
+          cache: 'pip'
       - name: Install deps
         run: |
           sudo snap install charmcraft --classic
           python -m pip install --upgrade pip
-          pip install tox pip-tools wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -24,13 +18,13 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Cache pip
-        uses: actions/cache@v2
+        uses: actions/cache@v3
```
mateoflorido, line 27: What if we switch to the [cache property](https://github.com/actions/setup-python#caching-packages-dependencies) bundled in `setup-python`?

addyess: what's odd here is that we're running all the tests with tox... so all the caching in the world doesn't really help with the fact that tox goes and recreates it's own virtual environment.  It does beg the question though maybe the github actions could run without tox?

mateoflorido: You're right, caching doesn't really help. Concerning the running the Github actions without tox from my point of view, this change would not reduce our testing time enough to undertake those changes

addyess: let's try it anyway 👍🏻

addyess: @mateoflorido cool... please re-review.  Looks like this idea works

mateoflorido: Nice! LGTM, I'll rerun the tests to see if the pip cache kicks in

mateoflorido: Great, it looks like it's working and it's faster as well. 🥇

mateoflorido (approved):
LGTM! 🥇

PR #1105, opened 2022-11-21
build containerd resources with charm builds

addyess:
Upon merging this PR, containerd will need to attach the 1.6.10 resource to the charm published to charmhub
*  https://github.com/charmed-kubernetes/charm-containerd/pull/84


addyess:
Nevermind, i think we'll manage this resource out of band of nightly charm builds

PR #1109, opened 2022-11-23
failed juju deployments should report failures to show up in generated reports

addyess:
Before exiting and moving on the cleanup stage, a failed deployment ought to post that it's failed its deployment so that the run will show up as a failure in the [generated reports](https://github.com/charmed-kubernetes/jenkins/blob/main/bin/report#L167-L173)

The report job was filtering out results that don't include a `result` metadata

### DIFF
--- a/ci.bash
+++ b/ci.bash
@@ -194,11 +194,6 @@ function ci::run
         juju::deploy
         juju::wait
 
-        kv::set "deploy_result" "True"
-        kv::set "deploy_endtime" "$(timestamp)"
-        touch "meta/deployresult-True"
-        python bin/s3 cp "meta/deployresult-True" "meta/deployresult-True"
-
         juju::deploy::after
 
         test::execute result
--- a/juju.bash
+++ b/juju.bash
@@ -86,14 +86,19 @@ function juju::wait
     timeout 45m juju-wait -e "$JUJU_CONTROLLER:$JUJU_MODEL" -w
 
     ret=$?
+
+    is_pass="True"
     if (( ret > 0 )); then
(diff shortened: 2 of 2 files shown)


ci.bash
```diff
@@ -194,11 +194,6 @@ function ci::run
         juju::deploy
         juju::wait
 
-        kv::set "deploy_result" "True"
-        kv::set "deploy_endtime" "$(timestamp)"
-        touch "meta/deployresult-True"
-        python bin/s3 cp "meta/deployresult-True" "meta/deployresult-True"
-
```
addyess, line 201: pulling this into `juju::wait`

juju.bash
```diff
@@ -86,7 +86,19 @@ function juju::wait
...
+    is_pass="True"
+    if (( ret > 0 )); then
+        is_pass="False"
+    fi
+
+    kv::set "deploy_result" "$(is_pass)"
+    kv::set "deploy_endtime" "$(timestamp)"
+    touch "meta/deployresult-$(is_pass)"
+    python bin/s3 cp "meta/deployresult-$(is_pass)" "meta/deployresult-$(is_pass)"
+
```
addyess, line 99: setting `deployresult` metadata based on the success of `juju-wait`

juju.bash
```diff
@@ -86,7 +86,19 @@ function juju::wait
...
     if (( ret > 0 )); then
+        test::report "False"
```
addyess, line 101: indicating the test failed if `juju-wait` fails

stonepreston (approved):

PR #1111, opened 2022-11-23
add try/finally blocks so ceph tests cleanup better.

addyess:


mateoflorido (approved):
LGTM!

jobs/integration/validation.py
```diff
@@ -2443,73 +2443,75 @@ async def test_ceph(model, tools):
...
+            "ceph-fs",
+            num_units=1,
+            series=series,
+            config=ceph_config,
+            channel=ceph_charms_channel,
+        )
 
-    # until bug https://bugs.launchpad.net/charm-kubernetes-control-plane/+bug/1824035 fixed
-    unit = model.applications["ceph-mon"].units[0]
-    await juju_run_action(unit, "create-pool", name="ext4-pool")
```
addyess, line 2487: removing this because the control-plane charm does this just fine.

jobs/integration/utils.py
```diff
@@ -343,41 +343,38 @@ async def log_snap_versions(model, prefix="before"):
...
 kind: PersistentVolumeClaim
 metadata:
-  name: {0}-pvc
-  annotations:
-   volume.beta.kubernetes.io/storage-class: {0}
+  name: {sc_name}-pvc
 spec:
   accessModes:
   - ReadWriteOnce
+  storageClassName: {sc_name}
```
addyess, line 354: quoting https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#using-dynamic-provisioning

> Users request dynamically provisioned storage by including a storage class in their PersistentVolumeClaim. Before Kubernetes v1.6, this was done via the volume.beta.kubernetes.io/storage-class annotation. However, this annotation is deprecated since v1.9. Users now can and should instead use the storageClassName field of the PersistentVolumeClaim object. The value of this field must match the name of a StorageClass configured by the administrator

PR #1118, opened 2022-11-30
multus validation: adjust deployment of ceph-osd and ceph-mon for jammy

addyess:
* multus and sriov specifications both need ceph as a k8s storage platform
* don't need to run the ceph-test for these two job specification (it would remove ceph)
* refactor a pytest fixture `skip_by_app` to handle an application predicate

if a test is skipped with this fixture, it will look like this in the log

```
jobs/integration/validation.py::test_ceph SKIPPED ('lambda apps: all(a in apps for a in ["ceph-mon", "ceph-osd"])' was True)
```

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -318,12 +318,14 @@ async def proxy_app(model):
 
 @pytest.fixture(autouse=True)
 def skip_by_app(request, model):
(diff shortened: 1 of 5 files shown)


stonepreston (approved):

addyess:
thanks @stonepreston

PR #1120, opened 2022-12-01
improve juju controller bootstrap times on vsphere

addyess:
by using the built-in templates on vsphere, we can bootstrap juju controllers much quicker


mateoflorido (approved):
LGTM

PR #1126, opened 2022-12-06
Update tag format in CAPI release

sachinkumarsingh092:
Updating the tag description to use the correct version format when doing a CAPI release.


ktsakalozos (approved):
LGTM +1

PR #1131, opened 2022-12-06
improve reporting of failing calico tests (and address some test failures)

addyess:


jobs/validate/calico-spec
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # shellcheck disable=SC2034,SC1090
 
-set -ex
+set -x
```
Cynerva, line 4: This change seems kinda scary. It means that any of these juju or tigera_aws.py commands could fail and we would just... ignore the error and move on, right?

addyess: hmm... well, yes that's true.  But I imagine it would result in a later test failure.  The current situation is that if any of those commands fail, the only commands that execute post failure are the `ci::cleanup` trap in ci.bash. 

This results in a cleaned up model but no test reporting on jenkaas.

addyess: maybe there's a good way to "wrap and test" all the potential failures better than `set -e` in the whole file

Cynerva: Yeah, I guess this is better than how it was. Better to have a slow test run and a confusing failure report than to have no report at all.

I do think it would be better if we could find a way to fix this while preserving the fail-fast nature of the script, but it seems like these jobs and the report handling aren't really set up to do it well. I think we would either have to keep `set -e` and report run failures in a trap, or remove `set -e` and add a lot of return code checks to the script. Or rewrite the whole thing in Python.

Cynerva (approved):
Approving this despite my concerns about error handling. I do think it's an improvement over how it was before.

jobs/validate/calico-spec
addyess: i went with the approach of wrapping the "must work" methods with `( ... )` and then checking the whole exit code of that bash block.  If there is an error, we'll call that a "failed deployment"

jobs/validate/calico-spec
```diff
@@ -88,13 +93,19 @@ function juju::deploy
...
 
-    if [ "$ROUTING_MODE" = "bgp-router" ]; then
-      python $WORKSPACE/jobs/integration/tigera_aws.py configure-bgp
+      if [ "$ROUTING_MODE" = "bgp-router" ]; then
+        python $WORKSPACE/jobs/integration/tigera_aws.py configure-bgp
+      fi
+    )
+    ret=$?
+    if (( ret > 0 )); then
+        juju:Ldeploy-failure $ret
```
Cynerva, line 108:
```suggestion
        juju:deploy-failure $ret
```

jobs/validate/calico-spec
```diff
@@ -88,13 +93,19 @@ function juju::deploy
...
          --channel "$JUJU_DEPLOY_CHANNEL" "$JUJU_DEPLOY_BUNDLE"
 
-    if [[ "$ROUTING_MODE" = bgp* ]]; then
-      python $WORKSPACE/jobs/integration/tigera_aws.py disable-source-dest-check
-      python $WORKSPACE/jobs/integration/tigera_aws.py assign-ipv6-addrs
-    fi
+    (
+      if [[ "$ROUTING_MODE" = bgp* ]]; then
+        python $WORKSPACE/jobs/integration/tigera_aws.py disable-source-dest-check
+        python $WORKSPACE/jobs/integration/tigera_aws.py assign-ipv6-addrs
```
Cynerva, line 99: Looks like errors from these are still ignored, since the subshell will only return the code of the last command run within it. e.g.

```
$ (false; true)
$ echo $?
0
```

Maybe a `set -e` in the subshell would do what's intended here?

```
(
  set -e
  ...
)
ret=$?
```

jobs/validate/calico-spec
Cynerva: That sounds good to me. Thanks.

Cynerva (approved):

PR #1132, opened 2022-12-07
Validation/ck8s localhost

addyess:
Make localhost lxd testing of Charmed-Kubernetes weekly rather than daily

- [ ] Needs `jjb` after merge


kwmonroe (approved):
heh, this is good.  totally unnecessary because we have better recommendations, but by golly, this is good.

kwmonroe:
needs a jjb on validate.yaml

PR #1133, opened 2022-12-07
[Reporting] successfully deployed tests should report such

addyess:
This is a partial reversion of https://github.com/charmed-kubernetes/jenkins/pull/1131

It's important that the `meta/deployresult-<True|False>` file exist in the archive.  Without it, failed tests will appear as "red' looking like they failed to deploy rather than "yellow" where they just fail a test"

### DIFF
--- a/jobs/validate/calico-spec
+++ b/jobs/validate/calico-spec
@@ -43,7 +43,8 @@ function juju::bootstrap
 
       ret=$?
       if (( ret > 0 )); then
-          juju::deploy-failure $ret
+          # Fail Deploy Early
+          juju::deploy-report $ret
       fi
     fi
 }
@@ -104,10 +105,7 @@ function juju::deploy
         python $WORKSPACE/jobs/integration/tigera_aws.py configure-bgp
       fi
     )
-    ret=$?
-    if (( ret > 0 )); then
-        juju::deploy-failure $ret
-    fi
+    juju::deploy-report $?
 }
 
 ###############################################################################
--- a/jobs/validate/localhost-spec
+++ b/jobs/validate/localhost-spec
@@ -26,11 +26,7 @@ function juju::deploy
         juju relate ubuntu ci-setup
     )
 
-    ret=$?
-    if (( ret > 0 )); then
-        juju::deploy-failure $ret
-    fi
-
+    juju::deploy-report $?
 }
 
 function juju::deploy::after
--- a/juju.bash
+++ b/juju.bash
@@ -75,10 +75,7 @@ function juju::deploy
          --force \
          --channel "$JUJU_DEPLOY_CHANNEL" "$JUJU_DEPLOY_BUNDLE"
 
-    ret=$?
-    if (( ret > 0 )); then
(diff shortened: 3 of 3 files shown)


juju.bash
```diff
@@ -75,10 +75,7 @@ function juju::deploy
          --force \
          --channel "$JUJU_DEPLOY_CHANNEL" "$JUJU_DEPLOY_BUNDLE"
 
-    ret=$?
-    if (( ret > 0 )); then
-        juju::deploy-failure $ret
-    fi
+    juju::deploy-report $?
```
addyess, line 78: this re-introduces previous behavior deleted [here](https://github.com/charmed-kubernetes/jenkins/pull/1131/files#diff-1c740f8e2163fd9b9ae38b82dbe468314d7384fb48606010c0a89d9833c3f024L96-L99)

jobs/validate/calico-spec
```diff
@@ -43,7 +43,8 @@ function juju::bootstrap
 
       ret=$?
       if (( ret > 0 )); then
-          juju::deploy-failure $ret
+          # Fail Deploy Early
+          juju::deploy-report $ret
```
addyess, line 47: a special case to fail fast on calico deployments

jobs/validate/calico-spec
```diff
@@ -104,10 +105,7 @@ function juju::deploy
         python $WORKSPACE/jobs/integration/tigera_aws.py configure-bgp
       fi
     )
-    ret=$?
-    if (( ret > 0 )); then
-        juju::deploy-failure $ret
-    fi
+    juju::deploy-report $?
```
addyess, line 108: report whether calico deployed (if it failed -- exit)

jobs/validate/localhost-spec
```diff
@@ -26,11 +26,7 @@ function juju::deploy
         juju relate ubuntu ci-setup
     )
 
-    ret=$?
-    if (( ret > 0 )); then
-        juju::deploy-failure $ret
-    fi
-
+    juju::deploy-report $?
```
addyess, line 29: report whether localhost tests deployed (if it failed -- exit)

juju.bash
addyess: always REPORT if the deployment succeeded or not...

juju.bash
```diff
@@ -106,16 +103,22 @@ print(unit[units[0]]['public-address'])
...
     kv::set "deploy_endtime" "$(timestamp)"
     touch "meta/deployresult-${is_pass}"
     python bin/s3 cp "meta/deployresult-${is_pass}" "meta/deployresult-${is_pass}"
 
-    test::report ${is_pass}
-    exit "$ret"
+    if (( ret > 0 )); then
+        test::report ${is_pass}
+        exit $ret
+    fi
```
addyess, line 123: if we failed deployment, `exit` to fast fail the testing

mateoflorido (approved):
LGTM. This would allow us to get better reports on Jenkaas. Thanks 🌟

PR #1139, opened 2022-12-08
[Parameterize Validation Tests] Roll-up all validation jobs targetting stable or edge charms

addyess:
This will add a parameter with a default value to all the validation tests if they are matrix-style tests or not
* 1.26/stable -- charms-stable
* 1.27/edge -- charms-edge

This make is a one stop shop to update those defaults, as well as allowing someone to run a job like `validate-ck-stable` with either `1.26/stable` or `1.25/stable` charm versions

Because we do not have the [dynamic-axis](https://plugins.jenkins.io/dynamic-axis/) plugin installed, we can't make matrix values parameters :(  Perhaps this is request to IS to provide

- [ ] Needs `jjb` after merge


jobs/ci-master.yaml
```diff
@@ -65,6 +65,46 @@
...
+          name: snap_version
+          default: 1.26/stable
+          description: channel for charmed-kubernetes snap used in deployment
+
+- parameter:
+    name: snap-beta
+    parameters:
+      - string:
+          name: snap_version
+          default: 1.27/stable
```
stonepreston, line 97: should this be 1.27/beta?

addyess:
```suggestion
          default: 1.27/beta
```

stonepreston (approved):

PR #1142, opened 2022-12-09
1.26 ga tweaks

kwmonroe:
- Drop anything less than n-4 from our supported enums; upstream k8s doesn't release new things there, so there's no reason for us to keep checking those branches.

- Update the GA release doc with references to what we did for the 1.26 GA.

- Drive by to bump our validate-offline job to use `1.26/edge` (even though it's not a current job; it might become one in the future)


mateoflorido:
Small nit. What about the `Promote new K8s snaps` and `Run validate-charm-release-upgrade` tasks? There are a some `1.25` fragments there -- should we leave them as they are or change them to the current release version?

PR #1143, opened 2022-12-09
reduce false error detection when deploying coredns charm

addyess:
* allow this test to run on a gce cloud if you happen to have one lying around
* better handle transient `errors` in the coredns charm if they go away eventually
* use libjuju to remove coredns charm after it's completed its usefulness

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -1828,7 +1828,7 @@ async def test_encryption_at_rest(model, tools):
     assert b64_value not in result.output
 
 
-@pytest.mark.clouds(["ec2", "vsphere"])
+@pytest.mark.clouds(["ec2", "vsphere", "gce"])
 async def test_dns_provider(model, k8s_model, tools):
     control_plane_app = model.applications["kubernetes-control-plane"]
     control_plane_unit = control_plane_app.units[0]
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -1947,11 +1947,7 @@ async def get_offer():
             coredns = k8s_model.applications["coredns"]
 
         log("Waiting for CoreDNS charm to be ready")
-        while (
-            status := coredns.units[0].workload_status if coredns.units else None
-        ) != "active":
-            assert status not in ("blocked", "error")
-            await asyncio.sleep(5)
+        await k8s_model.wait_for_idle(raise_on_error=False, status="active")
```
mateoflorido, line 1950: I think we should add a comment here so that we can come back later when this issue is addressed.

addyess: i'm guessing the offensive (workaroundy) part here is the `raise_on_error=False`

this prevents the method from raising an exception when ANY error occurs.  Now, if the charm isn't active after the standard timeout (10m) then if it's still in error, a timeout error will occur

mateoflorido: Alright, that makes sense! :)

mateoflorido (approved):
LGTM 💯

PR #1148, opened 2022-12-12
1.26 (non-SQA) job updates

kwmonroe:
These changes won't affect SQA for 1.26, but we'll need to make them sooner or later:

- update conformance jobs
  - 1.26 by default
  - fix issue running `tox -- bash`; we could whitelist bash, but these jobs don't need to run with tox, so just bash 'em directly
  - jobs with this branch: https://jenkins.canonical.com/k8s/view/Conformance/

- update infra jobs
  - drop our dead agents (amd-0 and amd-1 were replaces by ps5-amd-*)
  - remove the golang apt pkg and ppa (we get go from the snap store)
  - include rust/cargo to build recent pip `cryptography` package; binary wheels are available for amd64, but we have to build this on other arches

- update release job
  - the job was manually updated/run in jenkins; need this PR to ensure the job source matches what we actually ran
---

- [x] Needs `jjb` after merge: arc-conformance.yaml, cncf-conformance.yaml, infra.yaml, release.yaml

### DIFF
--- a/jobs/arc-conformance.yaml
+++ b/jobs/arc-conformance.yaml
@@ -20,7 +20,7 @@
     parameters:
       - string:
           name: CK_VERSION
-          default: '1.25'
+          default: '1.26'
           description: |
             CK version to deploy. This will be used to set the snap track
             and to identify what k8s version is associated with the results.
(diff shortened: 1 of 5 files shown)


addyess (approved):
yeah, these are well reasoned

addyess (approved):
alrighty

PR #1149, opened 2022-12-13
remove CSIMigrationvSphere=True in 1.26.0 and beyond

addyess:
* remove in-tree vsphere testing at 1.26.0 and beyond
* similar to other cloud's [testing config](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/integration/templates/integrator-charm-data/gce/out-of-tree.yaml)

### DIFF
--- a/jobs/integration/templates/integrator-charm-data/vsphere/out-of-tree.yaml
+++ b/jobs/integration/templates/integrator-charm-data/vsphere/out-of-tree.yaml
@@ -14,3 +14,4 @@ in-relations:
     - kubernetes-control-plane:vsphere
   - - vsphere-integrator:clients
     - kubernetes-worker:vsphere
+in-tree-until: '1.25'


Cynerva:
Are we still going to have problems with the overlay in integrator-spec [here](https://github.com/charmed-kubernetes/jenkins/blob/805fd936f05022de64021c866237b586eac6ff1a/jobs/validate/integrator-spec#L37-L49)?

addyess:
> Are we still going to have problems with the overlay in integrator-spec [here](https://github.com/charmed-kubernetes/jenkins/blob/805fd936f05022de64021c866237b586eac6ff1a/jobs/validate/integrator-spec#L37-L49)?

I'll test this once the charms are created.

Cynerva:
OK, so if I'm reading right, this spec will first deploy Charmed Kubernetes with vsphere-integrator using the in-tree provisioner. Then it will run `test_integrator_charm.py`, which...

On k8s 1.25 and earlier: will test the in-tree provider, then upgrade to the out-of-tree provider and test that too
On k8s 1.26 and later: will skip testing the in-tree provider, but will still upgrade to the out-of-tree provider to test that

If so then I think once the CSIMigrationvSphere fixes land, then the changes here should make the tests pass. As long as nothing tries to create a PVC prior to the upgrade to the out-of-tree provider.

Cynerva (approved):

addyess:
And after testing locally, these tests passed with the new charms

PR #1153, opened 2022-12-31
isolate ps5-12 for juju3 CI testing

kwmonroe:
Hacky way to isolate one of our jenkins workers with juju 3.  This does a regular infra job for ps5-12, and then a `snap refresh juju` with whatever param we set for the job.

- [x] Needs `jjb` after merge: infra-ps5.yaml

If we like this approach, we need to temporarily remove all labels from [jenkins-slave-focal-ps5-12](https://jenkins.canonical.com/k8s/computer/jenkins-slave-focal-ps5-12/) except for `runner-ps5-amd64-12`.  Any job that needs to be run against juju3 will need to have their 'run on' label temporarily set to `runner-ps5-amd54-12`.


jobs/infra-ps5.yaml
```diff
@@ -55,3 +55,63 @@
...
+              [[ -f $WORKSPACE/.env ]] && source $WORKSPACE/.env
+              set +o allexport
+
+              bash jobs/infra/fixtures/cleanup-env.sh
+
+              venv/bin/pip install ansible
+              venv/bin/ansible-playbook jobs/infra/playbook-jenkins.yml --limit localhost --tags 'jenkins' -i jobs/infra/hosts
+
+              echo Overriding juju snap version
+              sudo snap refresh juju --channel $JUJU_CHANNEL
```
kwmonroe, line 117: this `snap refresh` after the playbook is the only difference between this job and the normal infra jobs.

jobs/infra-ps5.yaml
```diff
@@ -55,3 +55,63 @@
...
+      Temp job for testing juju3. This is just like the other infra jobs,
+      but exposes the ability to change the juju snap on the ps5-12 node.
+    node: runner-ps5-amd64-12
+    project-type: freestyle
+    scm:
+      - k8s-jenkins-jenkaas
+    parameters:
+      - string:
+          name: JUJU_CHANNEL
+          default: '3.0/candidate'
```
addyess, line 71: any reason to not use `3.1/stable` here?

addyess (changes requested):
so, the normal infra job installs `2.9/stable`

why are we picking 3.0/candidate?  Can we go with 3.1/candidate?

jobs/infra-ps5.yaml
kwmonroe: nah, no reason.  let's do it!

addyess (approved):
🍰

PR #1155, opened 2023-01-04
use the nagios host refresh api to encourage faster status checking

addyess:
using a POST API in nagios, one can request a refresh of all the services on a specific host.  If we do this, we can speed these tests up from taking 20-30 minutes to about 10-15 minutes.


mateoflorido (approved):
Nice, LGTM :)

PR #1158, opened 2023-01-06
migrate tests from multus integration tests into a conjoined ovn and multus validation tests

addyess:
Porting the integration tests from [multus charm](https://github.com/charmed-kubernetes/charm-multus/tree/main/tests/integration) to become the validation tests for the newly conjoined ovn and multus tests.  

- [ ] Needs `jjb` after merge for validation.yaml


jobs/integration/validation.py
```diff
@@ -2085,104 +2085,128 @@ async def test_multus(model, tools, addons_model):
...
+            f"/snap/bin/kubectl --kubeconfig /root/.kube/config apply -f {remote_path}; rm {remote_path}",
         )
-        await multus_app.set_config({"network-attachment-definitions": ""})
 
     await cleanup()
 
-    # Create NetworkAttachmentDefinition for Flannel
+    await containerd_app.set_config(
+        {
+            "no_proxy": "localhost,127.0.0.1,::1,10.246.154.0/24,10.152.183.0/24,192.168.0.0/16"
```
Cynerva, line 2116: This could cause problems if we ever want to run this test manually in environments other than our vSphere lab.

Could this proxy config be set in the deployment overlay in ovn-multus-spec instead? I think it would be nice if we can consolidate our environment-specific configuration there so that test code like this can be less destructive.

jobs/integration/validation.py
```diff
@@ -2085,104 +2085,128 @@ async def test_multus(model, tools, addons_model):
...
+            protocol="IPv4",
+            cidrBlock="10.166.0.0/16",
+            default=False,
+            excludeIps=["10.166.0.1"],
+            gateway="10.166.0.1",
+            gatewayType="distributed",
+            natOutgoing=True,
+            namespaces=["default"],
+        ),
+    )
```
Cynerva, line 2156: I'm a little puzzled that this subnet does not have the `provider` field set, which should be set to `attachnet.default` according to the Kube-OVN Multi-NIC docs [here](https://kubeovn.github.io/docs/v1.10.x/en/advance/multi-nic/#create-a-kube-ovn-subnet). But I see this subnet is copied from the Multus charm integration tests which haven't encountered any issues with the subnet, so... I guess it's not required? Seem weird, but, if it's working then I won't ask for any change here.

jobs/integration/validation.py
```diff
@@ -2085,104 +2085,128 @@ async def test_multus(model, tools, addons_model):
...
+                if line.startswith("inet "):
+                    # interface has an address, we'll call that good enough
+                    active_networks[interface] = line.split()[1]
+        expected_interfaces = ["eth0", "net1"]
+        for ifc in expected_interfaces:
+            assert (
+                ifc in active_networks
+            ), f"Interface {ifc} is missing from ip addr output:\n{output}"
+            assert active_networks[ifc].startswith(
+                "10.166.0."
```
Cynerva, line 2206: It's a `/16` subnet so only the first two octets are fixed

```suggestion
                "10.166."
```

Cynerva (changes requested):

Cynerva (approved):
LGTM

PR #1162, opened 2023-01-10
upgrade juju_wait() to support `tee`ing of stdout and stderr

addyess:
Many integration tests use `juju_wait` to wait for a juju model to be active.  Some of these tests don't specify a timeout period, so the tests fail after running for hours because the test is interrupted.  If interrupted, there's  not a debug log of what the model was waiting to complete.

This PR updates `juju_wait` calls to tee `juju-wait`'s  stderr to the debug log as well as still returning stdout and stderr to the caller.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -179,7 +179,7 @@ class Tools:
             [
                 _read_stream(proc.stdout, lambda l: tee(l, stdout, 1)),
                 _read_stream(proc.stderr, lambda l: tee(l, stderr, 2)),
-                _feed_stream(input=stdin)
+                _feed_stream(input=stdin),
             ]
         )
 
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -161,7 +161,7 @@ class Tools:
         if "m" not in kwargs:
             kwargs["m"] = self.connection
         debug = partial(click.echo, nl=False)
-        result = command("-w", "-v", *args, **kwargs, _out=debug, _tee=True)
+        result = command("-w", "-v", *args, **kwargs, _err=debug, _tee="err")
         stdout, stderr = result.stdout.decode("utf-8"), result.stderr.decode("utf-8")
         return stdout, stderr
 
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -164,16 +164,22 @@ class Tools:
 
         async def _read_stream(stream, callback):
             while True:
-                line = await stream.readline()
+                line = await stream.read()
                 if line:
                     callback(line)
                 else:
                     break
 
+        async def _feed_stream(input):
+            if input:
+                # replicates what proc.communicate() does with stdin
+                await proc._feed_stdin(input)
+
         await asyncio.wait(
             [
                 _read_stream(proc.stdout, lambda l: tee(l, stdout, 1)),
                 _read_stream(proc.stderr, lambda l: tee(l, stderr, 2)),
+                _feed_stream(input=stdin)
             ]
         )
 
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -153,10 +153,6 @@ class Tools:
     def juju_wait(self, *args, **kwargs):
         """Run juju-wait command with provided arguments.
 
-        method is defined as async -- but it isn't necessary
-        as the latest implementation uses `sh` and doesn't
-        support async/await.
-
         if kwarg contains `m`: juju-wait is executed on a different model
         see juju-wait --help for other supported arguments
         """
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -162,8 +162,7 @@ class Tools:
             kwargs["m"] = self.connection
         debug = partial(click.echo, nl=False)
         result = command("-w", "-v", *args, **kwargs, _err=debug, _tee="err")
-        stdout, stderr = result.stdout.decode("utf-8"), result.stderr.decode("utf-8")
-        return stdout, stderr
+        return result.stdout.decode("utf-8"), result.stderr.decode("utf-8")
 
 
 @pytest.fixture(scope="module")
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -164,7 +164,7 @@ class Tools:
 
         async def _read_stream(stream, callback):
             while True:
-                line = await stream.read()
+                line = await stream.read(1024)
                 if line:
                     callback(line)
                 else:
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -129,7 +129,19 @@ class Tools:
         self.is_series_upgrade = request.config.getoption("--is-series-upgrade")
         self.charm_channel = request.config.getoption("--charm-channel")
 
-    async def run(self, cmd, *args, stdin=None, _tee=False):
+    async def run(self, cmd: str, *args: str, stdin=None, _tee=False):
+        """
+        asynchronously run a command as a subprocess
+
+        @param str cmd: path to command on filesystem
+        @param str *args: arguments to the command
+        @param Optional[bytes] stdin: data to pass over stdin
+        @param _tee:
+            False -- neither stdout nor stderr is tee'd
+            True  -- stdout and stderr are both tee'd
+            "out" -- stdout is tee'd to test stdout
+            "err" -- stderr is tee'd to test stderr
+        """
         proc = await asyncio.create_subprocess_exec(
             cmd,
             *args,
@@ -144,11 +156,11 @@ class Tools:
 
         stdout, stderr = bytearray(), bytearray()
 
-        def tee(line:bytes, sink:bytearray, fd: int):
+        def tee(line: bytes, sink: bytearray, fd: int):
             sink += line
-            if _tee == "err" and fd==2:
-                os.write(fd, line)
-            if _tee in (True, "out") and fd==1:
+            write |= _tee == "out" and fd == 1
+            write |= _tee == "err" and fd == 2
+            if write or _tee == True:
                 os.write(fd, line)
 
         async def _read_stream(stream, callback):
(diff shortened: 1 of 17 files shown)


jobs/integration/conftest.py
```diff
@@ -1,19 +1,22 @@
...
 import yaml
-import requests
-import click
+
 from datetime import datetime
+from functools import partial
+from juju.model import Model
+from pathlib import Path
 from py.xml import html
 from tempfile import NamedTemporaryFile
```
addyess, line 19: Just rearranging the import order so that its easier to find included modules

jobs/integration/conftest.py
```diff
@@ -150,13 +150,24 @@ async def run(self, cmd, *args, stdin=None):
...
+        as the latest implementation uses `sh` and doesn't
+        support async/await.
+
+        if kwarg contains `m`: juju-wait is executed on a different model
+        see juju-wait --help for other supported arguments
+        """
+
+        command = sh.Command("/snap/bin/juju-wait")
+        if "m" not in kwargs:
+            kwargs["m"] = self.connection
```
addyess, line 166: `-e` and `-m` are the same command switch.  I've replaced with `m` to closer match `juju` command and to pick up certain `juju_wait(...)` calls that were supposed to be waiting on a different model

jobs/integration/conftest.py
```diff
@@ -150,13 +150,24 @@ async def run(self, cmd, *args, stdin=None):
...
+        debug = partial(click.echo, nl=False)
+        result = command("-w", "-v", *args, **kwargs, _out=debug, _tee=True)
+        stdout, stderr = result.stdout.decode("utf-8"), result.stderr.decode("utf-8")
+        return stdout, stderr
```
addyess, line 170: This is the significant chunk of changes

jobs/integration/conftest.py
```diff
@@ -150,13 +150,20 @@ async def run(self, cmd, *args, stdin=None):
...
+
+        if kwarg contains `m`: juju-wait is executed on a different model
+        see juju-wait --help for other supported arguments
+        """
+
+        command = sh.Command("/snap/bin/juju-wait")
+        if "m" not in kwargs:
+            kwargs["m"] = self.connection
+        debug = partial(click.echo, nl=False)
+        result = command("-w", "-v", *args, **kwargs, _err=debug, _tee="err")
```
addyess, line 164: tee`ing the stderr from `juju-wait` should provide realtime results from juju-wait as well as showing the response from juju-wait in the ci.log file

jobs/integration/conftest.py
```diff
@@ -150,13 +150,19 @@ async def run(self, cmd, *args, stdin=None):
...
         return stdout.decode("utf8"), stderr.decode("utf8")
 
-    async def juju_wait(self, *args, **kwargs):
-        cmd = ["/snap/bin/juju-wait", "-e", self.connection, "-w", "-v"]
-        if args:
-            cmd.extend(args)
-        if "timeout_secs" in kwargs and kwargs["timeout_secs"]:
-            cmd.extend(["-t", str(kwargs["timeout_secs"])])
-        return await self.run(*cmd)
+    def juju_wait(self, *args, **kwargs):
```
Cynerva, line 153: I'm not so sure about this. This is going to make every call to `juju_wait` block the current task without yielding to the event loop. Every task in the event loop will be blocked for potentially several minutes at a time.

I'm worried about background tasks getting blocked by this, particularly in python-libjuju which is heavily based in coroutines. See for example this [watcher task](https://github.com/charmed-kubernetes/python-libjuju/blob/fd94f0635a36518573a209a93f5361d3152c6df4/juju/model.py#L1260-L1264) that watches for model events to update model state.

addyess: alrighty -- i'll refactor to keep using the async... but still with tee'd stdout

jobs/integration/conftest.py
```diff
@@ -1,19 +1,23 @@
...
-import requests
-import click
+
 from datetime import datetime
+from functools import partial
+from juju.model import Model
+from pathlib import Path
 from py.xml import html
 from tempfile import NamedTemporaryFile
-from juju.model import Model
```
addyess, line 16: resorting the import order for finding modules alphabetically

jobs/integration/conftest.py
```diff
@@ -141,22 +154,52 @@ async def run(self, cmd, *args, stdin=None):
         if hasattr(stdin, "encode"):
             stdin = stdin.encode("utf8")
 
-        stdout, stderr = await proc.communicate(input=stdin)
-        if proc.returncode != 0:
```
addyess, line 145: rather than use `proc.communicate` which blocks until the process returns

jobs/integration/conftest.py
```diff
@@ -141,22 +154,52 @@ async def run(self, cmd, *args, stdin=None):
...
+        stdout, stderr = bytearray(), bytearray()
+
+        def tee(line: bytes, sink: bytearray, fd: int):
+            sink += line
+            write |= _tee == "out" and fd == 1
+            write |= _tee == "err" and fd == 2
+            if write or _tee == True:
+                os.write(fd, line)
```
addyess, line 164: append the subprocess `stdout` and `stderr` to a `bytesarray` object while writing to the test's `stdout` and `stderr`

jobs/integration/conftest.py
```diff
@@ -141,22 +154,52 @@ async def run(self, cmd, *args, stdin=None):
...
+                    callback(line)
+                else:
+                    break
+
+        await asyncio.wait(
+            [
+                _read_stream(proc.stdout, lambda l: tee(l, stdout, 1)),
+                _read_stream(proc.stderr, lambda l: tee(l, stderr, 2)),
+            ]
+        )
```
addyess, line 179: while reading from the process' streams

jobs/integration/conftest.py
```diff
@@ -141,22 +154,52 @@ async def run(self, cmd, *args, stdin=None):
...
+
+        return_code = await proc.wait()
```
addyess, line 181: and retrieve the return_code from `proc.wait()`

jobs/integration/conftest.py
```diff
@@ -141,22 +154,52 @@ async def run(self, cmd, *args, stdin=None):
...
+
+        if kwargs contains `m`: juju-wait is executed on a different model
+        if kwargs contains `max_wait`: a timeout value is set
+        see juju-wait --help for other supported arguments
+        """
+        if "m" not in kwargs:
+            kwargs["m"] = self.connection
+        kwargs.update(dict(w=True, v=True))  # workload + verbose
+        juju_wait = sh.Command("/snap/bin/juju-wait")
+        command = shlex.split(str(juju_wait.bake(**kwargs)))
```
addyess, line 201: use `sh` to assemble the command which is passed to the async runner.

jobs/integration/conftest.py
```diff
@@ -141,22 +153,52 @@ async def run(self, cmd, *args, stdin=None):
...
-        stdout, stderr = await proc.communicate(input=stdin)
-        if proc.returncode != 0:
+        stdout, stderr = bytearray(), bytearray()
+
+        def tee(line: bytes, sink: bytearray, fd: int):
+            sink += line
+            write = _tee == "out" and fd == 1
+            write |= _tee == "err" and fd == 2
+            if write or _tee is True:
+                os.write(fd, line)
```
addyess, line 163: append the subprocess `stdout` and `stderr` to a `bytesarray` object while writing to the test's `stdout` and `stderr`

jobs/integration/conftest.py
Cynerva: It looks like we're waiting here for the stdout and stderr streams to end before moving on. But, we haven't sent stdin to the process. Actually, it looks like we never do that?

Indeed, if I make this a standalone function and run it like so:
```
out, err = await run("cat", stdin="foo")
```

the call hangs forever, because `cat` is waiting for stdin that never gets sent to it.

I think this would need a `_write_stream` or something, too

addyess: i tested with this:

```python
async def test_tools_run(tools):
    out, err = await tools.run("cat", stdin="foo", _tee=True)
    assert out == "foo"
    assert err == ""
  ```

File: jobs/integration/conftest.py [hunk 13]
```diff
@@ -141,22 +153,58 @@ async def run(self, cmd, *args, stdin=None):
...
+            if input:
+                # replicates what proc.communicate() does with stdin
+                await proc._feed_stdin(input)
+
+        await asyncio.wait(
+            [
+                _read_stream(proc.stdout, lambda l: tee(l, stdout, 1)),
+                _read_stream(proc.stderr, lambda l: tee(l, stderr, 2)),
+                _feed_stream(input=stdin)
+            ]
```
addyess, line 183: this is essentially what's done in [asyncio.subprocess](https://github.com/python/cpython/blob/3.11/Lib/asyncio/subprocess.py#L182-L197)

jobs/integration/conftest.py
```diff
@@ -141,22 +153,58 @@ async def run(self, cmd, *args, stdin=None):
...
+        def tee(line: bytes, sink: bytearray, fd: int):
+            sink += line
+            write = _tee == "out" and fd == 1
+            write |= _tee == "err" and fd == 2
+            if write or _tee is True:
+                os.write(fd, line)
+
+        async def _read_stream(stream, callback):
+            while True:
+                line = await stream.read()
```
Cynerva, line 167: This blocks until EOF is received. So the output doesn't get tee'd until the subprocess closes stdout/stderr at its end. I'm assuming we want to see tee'd output before then.

See: [StreamReader.read](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.read). I think we need to pass a number to prevent it from waiting for EOF. In my testing with `read(1024)`, it blocks until at least 1 byte is received and then returns however many bytes are available, up to 1024. I think that should work for our purposes.

```suggestion
                line = await stream.read(1024)
```

Cynerva (approved):

PR #1182, opened 2023-01-24
Improve bugfix release documentation

addyess:
Will need `jjb` after merging to pick up changes to the `promote-charms` job. 

This provides the release manager better clarity of the parameters to use when running the promote charms job


mateoflorido (approved):
LGTM! Thanks @addyess ✨

PR #1188, opened 2023-01-30
reintroduce snap revisions to support in sync-snaps

addyess:


kwmonroe:
I don't *think* we need this anymore.  This bit us because we immediately stopped looking for `n-3` point-releases when `ck-1.n` went GA.  Since we only support `n-2`, we would miss `n-3` upstream tags.  There's a period of time after `n` is GA that `n-3` is still supported upstream (the [2-month maintenance period](https://kubernetes.io/releases/patch-releases/#support-period)).

I double checked upstream tags >= 1.16 and we have snaps for all of them. As long as we keep the `K8S_STARTING_SEMVER` at `n-3`, we should catch 'em all as we roll through new releases.

Reopen if I'm missing something, but I'd *really* like to not check every upstream tag >= 1.16 against our snap revs for all arches; the `sync-snaps` job would take forever.

kwmonroe:
Aight, this *was* caused by an n-4 eks request, but it was for 1.22.17. Here's how it went down:
```
dec-08
02:00 sync-snaps ran; latest 1.22 branch is 1.22.16
05:41 upstream tags/releases 1.22.17
13:51 upstream tags/releases 1.26.0

dec-09
02:00 sync-snaps fails; latest 1.22 branch remains at 1.22.16
08:44 enums are set for the ck 1.26 GA; K8S_STARTING_SEMVER is now 1.23.0

dec-10
02:00 sync-snaps runs; latest 1.22 branch remains at 1.22.16 (1.22 is now < K8S_STARTING_SEMVER)

jan-04
04:35 cpc requests 1.22.17
11:49 build-eks: ERROR: v1.22.17 branch not found
```
And then @addyess went off and did magic to get the train back on the tracks.  This was an unfortunate set of circumstances that got us out-of-sync with an n-4 release. Not only was the timing just right so that we moved our enums whilst not being in-sync with 1.22 tags, but upstream made the 1.22.17 tag [post EOL](https://kubernetes.io/releases/patch-releases/#non-active-branch-history).

All that to say i'm reopening this, but only going back n-4 for the K8S_STARTING_SEMVER.  I'm still reluctant to go back any farther because (a) there's nothing back there that we don't already have, and (b) sync-snaps would take forever.

Also @Cynerva wants the K8S_GO_MAP, so let's keep the dicts around for reference.

kwmonroe (approved):

PR #1193, opened 2023-02-03
bump stable releases for validate-ck-upgrade tests

addyess:
- [ ] Needs `jjb` after merge

### DIFF
--- a/jobs/validate.yaml
+++ b/jobs/validate.yaml
@@ -199,6 +199,7 @@
           values:
             - 1.26/stable
             - 1.25/stable
+            - 1.24/stable
       - axis:
           type: user-defined
           name: series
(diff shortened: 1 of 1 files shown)


Cynerva (approved):
It looks to me like this job deploys `1.26/stable` charms, with snap channel set to `1.24/stable` or `1.25/stable`. Then, it upgrades the charms to `latest/edge`. Then, it upgrades snaps to `1.26/stable`.

I would think that we do want to test upgrading to `latest/edge` from a charm that started with `1.26/stable` snaps, so adding `1.26/stable` here makes sense to me.

I'm not 100% sure about dropping `1.24/stable` since historically, our stance was that a given charm release should support k8s versions as far back as N-2, so `1.24/stable` snaps might still be a valid case we would want to test. It's not really clear these days if we still maintain that stance given that we now have charm tracks that we can release to independently. I guess I'm not too worried about it, so, LGTM.

Cynerva (approved):

PR #1197, opened 2023-02-07
Crash if k8s version is missing from K8S_GO_MAP

Cynerva:
I get nervous that we're gonna accidentally build something with the wrong golang version because a k8s version is missing from K8S_GO_MAP. I propose instead of continuing with a potentially wrong golang version, just crash instead.


kwmonroe (approved):

addyess (approved):
Fear no more

PR #1203, opened 2023-02-10
Update AWS iam testing to be re-runnable

addyess:
Monthly tests of aws-iam have been failing for a while due to [kubernetes deprecations](https://github.com/charmed-kubernetes/charm-aws-iam/pull/12)

These changes allowed me to re-run the test over and over without tearing down an already installed aws-iam-charm which was in the model. 

note the new error when trying to authenticate is `accessdenied` rather than `error` or `forbidden`


kwmonroe (approved):
:100:

PR #1206, opened 2023-02-10
Tested basic validation on vsphere with juju 3.0/stable channel

addyess:
This PR allows for validation testing to run with a juju 2.9 or juju 3.0 client/controller pair.  
The snap which is used to bootstrap the controller, should be the same snap version used while running the tests. 

There's even a bit of preparation in this PR for juju 3.1 though it's not yet fully tested

Other tests may still be failing with juju 3.0 -- only the basic ck-validation on vsphere was tested

Depends on the successful merge and  build of a kubernetes-control-plane charm 
* https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/267

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -149,7 +149,7 @@ class Tools:
             "focal": "ubuntu@20.04",
             "jammy": "ubuntu@22.04",
         }
-        return f"--series={mapping[series]}"
+        return f"--base={mapping[series]}"
 
     async def run(self, cmd: str, *args: str, stdin=None, _tee=False):
         """
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -122,6 +122,9 @@ class Tools:
     async def _load(self):
         request = self._request
         whoami, _ = await self.run("juju", "whoami", "--format=yaml")
+        stdout, _ = await self.run("juju", "--version")
+        ver_str = stdout.splitlines()[-1].split("-")[0]
+        self.juju_version = tuple(map(int, ver_str.split(".")))
         self.juju_user = yaml.safe_load(whoami)["user"]
         self.controller_name = request.config.getoption("--controller")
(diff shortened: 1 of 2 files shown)


kwmonroe (approved):
LGTM, clever handling for `series` changes.  I'm guessing we'll find lots of places where we need to refactor `--series`.

mateoflorido (approved):
LGTM2

jobs/integration/conftest.py
```diff
@@ -137,6 +140,17 @@ async def _load(self):
...
+    def juju_base(self, series):
+        """Retrieve juju 3.1 base from series."""
+        if self.juju_version < (3, 1):
+            return f"--series={series}"
+        mapping = {
+            "bionic": "ubuntu@18.04",
+            "focal": "ubuntu@20.04",
+            "jammy": "ubuntu@22.04",
+        }
+        return f"--series={mapping[series]}"
```
addyess, line 152: i need to look at this again, it may be
```suggestion
        return f"--base={mapping[series]}"
```

jobs/integration/conftest.py
```diff
@@ -137,6 +140,17 @@ async def _load(self):
...
+    def juju_base(self, series):
+        """Retrieve juju 3.1 base from series."""
+        if self.juju_version < (3, 1):
+            return f"--series={series}"
+        mapping = {
+            "bionic": "ubuntu@18.04",
+            "focal": "ubuntu@20.04",
+            "jammy": "ubuntu@22.04",
+        }
+        return f"--base={mapping[series]}"
```
mateoflorido, line 152: I think '--base' is the one we need. However, after reading the 'juju deploy' [page](https://juju.is/docs/olm/juju-deploy), it seems that '--base' does something else (Existing storage to attach to the deployed unit). We should watch out for any unusual issues, but it's likely just a documentation issue.

PR #1213, opened 2023-02-17
Add validation --juju-ssh-proxy option

Cynerva:
Fixes https://bugs.launchpad.net/bugs/2007708

This adds a `--juju-ssh-proxy` option for validation that will proxy Juju SSH/SCP commands through the Juju controller. This should make it possible to run validation tests in situations where kubernetes-control-plane does not have a public IP.

### DIFF
--- a/jobs/integration/charm/test_etcd.py
+++ b/jobs/integration/charm/test_etcd.py
@@ -98,7 +98,11 @@ async def test_snapshot_restore(model, tools):
                 src = Path(action.results["snapshot"]["path"])
                 dst = Path(action.results["snapshot"]["path"]).name
                 await unit.scp_from(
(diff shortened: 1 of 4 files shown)


Basdbruijne (approved):

Cynerva:
I tested this with the Juju 2.9 CLI, but haven't tested with Juju 3.0/3.1. I'll give that a try.

addyess (approved):
Nice, don't hold it up if it doesn't work with juju 3.0

Cynerva:
Thanks folks. Strict confinement issues aside, I've verified at least that `juju scp --proxy` commands still work in Juju 3.1, so this should be a viable strategy for us going forward.

PR #1214, opened 2023-02-17
improvements to sync_all_track_snaps

kwmonroe:
- Move the pre-release and branch_version logic out of the $arch loop since our snap branches aren't arch-specific.
- Our snap recipe builds for all arches. Once we request a new one, break out of the $arch loop so we don't ask for more.


kwmonroe:
CI from this branch looks good:

https://jenkins.canonical.com/k8s/view/Sync/job/sync-snaps/943/console

addyess (approved):
I accept that the CI looks good,
I accept that code also is readable and maintain able.  --  much better than the previous `exclude_pre` statefulness in that loop

PR #1215, opened 2023-02-21
Set variables to create a new cluster for in-place upgrades

sachinkumarsingh092:
### Summary

For testing in-place e2e tests, we need to create new variables to create a new cluster to test the upgrades.


ktsakalozos (approved):
LGTM +1

PR #1216, opened 2023-02-21
Propagate changes in the `/root/cdk/serviceaccount.key` file

addyess:
Due to changes in which files are propagated by the k-c-p leader, to  non-leaders our validation testing needs to adjust which file it tests

* https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/pull/270

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -218,8 +218,10 @@ async def test_auth_file_propagation(model, tools):
         "md5sum /root/cdk/serviceaccount.key",
     )
 
+    # fail the test if we're waiting more than 10 minutes
+    max_wait = 10 * 60
+
     # Check that md5sum on non-leader matches
-    max_wait = 120
     follower_md5 = await run_until_success(
         follower,
         'md5sum /root/cdk/serviceaccount.key | grep "{}"'.format(leader_md5),
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -196,14 +196,14 @@ async def reset_audit_config(control_plane_app, tools):
 
 # START TESTS
 async def test_auth_file_propagation(model, tools):
-    """Validate that changes to /root/cdk/basic_auth.csv on the leader master
-    unit are propagated to the other master units.
+    """Validate that changes to /root/cdk/serviceaccount.key on the leader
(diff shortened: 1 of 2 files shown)


jobs/integration/validation.py
```diff
@@ -196,36 +196,42 @@ async def reset_audit_config(control_plane_app, tools):
...
-        "echo test,test,test >> /root/cdk/basic_auth.csv && "
-        "md5sum /root/cdk/basic_auth.csv",
+        "echo '#EXTRA' >> /root/cdk/serviceaccount.key && "
+        "md5sum /root/cdk/serviceaccount.key",
     )
 
     # Check that md5sum on non-leader matches
-    await run_until_success(
-        follower, 'md5sum /root/cdk/basic_auth.csv | grep "{}"'.format(leader_md5)
+    max_wait = 120
```
Cynerva, line 222: I'm a little worried that this timeout is too low, given that kubernetes-control-plane hooks usually take longer than a minute to run. A poorly timed update-status hook could prevent a leader-settings-changed hook from writing the new files quick enough. I would be tempted to put this at 5-10 minutes instead.

This is just gut feeling though. I'd be happy to try this at 120 and see what happens.

Cynerva (approved):

jobs/integration/validation.py
addyess: i've got 5 min to burn if you do

jobs/integration/validation.py
```diff
@@ -196,36 +196,44 @@ async def reset_audit_config(control_plane_app, tools):
...
     leader_md5 = await run_until_success(
         leader,
-        "echo test,test,test >> /root/cdk/basic_auth.csv && "
-        "md5sum /root/cdk/basic_auth.csv",
+        "echo '#EXTRA' >> /root/cdk/serviceaccount.key && "
+        "md5sum /root/cdk/serviceaccount.key",
     )
 
+    # fail the test if we're waiting more than 10 minutes
+    max_wait = 10 * 60
```
kwmonroe, line 222: good to see a ceiling on this!

kwmonroe (approved):
much better -- even the parts not related to the filename change.  nice.

PR #1224, opened 2023-02-27
keep 10 runs worth of job history

kwmonroe:
Keep 10 runs worth of job history.

---

- [x] Needs `jjb` after merge: affects almost all our jobs, so a pretty big jjb is needed.


addyess (approved):
10's a good number to me

PR #1227, opened 2023-03-09
ensure correct charm-matrix data

kwmonroe:
Partial fix for https://github.com/charmed-kubernetes/kubernetes-docs/issues/754

- fix incorrect labels (e.g. coredns is not a cni)
- ensure `k8s-operator` for all our sidecar charms
- add `docs-extra` where needed to populate the [components page](https://ubuntu.com/kubernetes/docs/1.26/components#additional-charms)
- add k8s-dashboard; we don't build/maintain this charm yet, but reasonable people would expect this to be listed as a ck8s "additional charms"

### DIFF
--- a/jobs/ci-master.yaml
+++ b/jobs/ci-master.yaml
@@ -218,7 +218,7 @@
           description: Channel to publish charm from (only used in promote jobs)
       - string:
           name: FILTER_BY_TAG
-          default: 'k8s'
+          default: 'k8s,k8s-operator'
           description: |
             Filter the builds by tag (ie. k8s). A tag can also be the name of a
             charm you want to individually build.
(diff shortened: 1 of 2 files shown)


addyess:
my understanding is that the daily charm builds are only for charms that are marked as with the `k8s` tag.  Perhaps we should update the  `FILTER_BY_TAG` parameter in [ci-master.yaml](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/ci-master.yaml#L221) to be `k8s,k8s-operator` so we consistently build these?

addyess (approved):
Yeah, i've re-read this a few times over and think these are sane changes.

kwmonroe:
> Perhaps we should update the `FILTER_BY_TAG` parameter in [ci-master.yaml](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/ci-master.yaml#L221) to be `k8s,k8s-operator` so we consistently build these?

Agreed and done. I considered just overriding the param in `build-charms.yaml`, but that's the only job that references the FILTER_BY param, so a ci-master change feels fine.

addyess (approved):
LGTM

PR #1239, opened 2023-03-20
destroy instances created by jenkins jobs, but not cleanly destroyed

addyess:


Cynerva:
I would put an explicit check somewhere in here that the vmfolder is `k8s-ci-root`. An accidental deletion of all VMs in any other folder could be catastrophic.

PR #1249, opened 2023-03-27
Refresh pip reqs

kwmonroe:
We're pretty far behind on our dependabot req.txt updates. Here's a PR that brings us all the way up2date with `pip-compile` in a py38 env.  I used py38 because our jenkins workers are 20.04 with py38.  One notable change is the ansible jump to v6, which should resolve infra warnings like this, caused by an old ansible + new jinja2:
```
11:19:30 [WARNING]: Skipping plugin (/var/lib/jenkins/slaves/jenkins-slave-
11:19:30 focal-3/workspace/infra-maintain-nodes-amd64-3/venv/lib/python3.8/site-
11:19:30 packages/ansible/plugins/filter/mathstuff.py) as it seems to be invalid: cannot
11:19:30 import name 'environmentfilter' from 'jinja2.filters'
```

This doesn't update the `requirements-2.9.txt`, which is used when testing against juju 2.9 envs.  We should bring that in-sync if this PR doesn't cause horrible issues.


kwmonroe:
Too risky for now.  We'll come back after 1.27 goes out the door.

PR #1252, opened 2023-03-27
remove sh output chained to other sh commands in git.py

addyess:
* `cilib/git.py` doesn't need to chain sh.output back through the shell to sort

### DIFF
--- a/cilib/git.py
+++ b/cilib/git.py
@@ -3,12 +3,17 @@
 
 import logging
 import sh
+import re
 from subprocess import run
 from .github_api import Repository
 import requests
 import requests.auth
 
 log = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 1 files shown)


kwmonroe:
Prereq for the `sh` update that will come with #1249

addyess:
@kwmonroe i've confirmed this works with both new and old versions of `sh`
Secondarily, it yields the same results as the previous "sh-pipe-python-thing"

```bash
git ls-remote -t --refs git+ssh://k8s-team-ci@git.launchpad.net/k8s-internal-mirror | sort -t / -k 3 -V | sed -E 's/^[[:xdigit:]]+[[:space:]]+refs\/tags\/(.+)/\1/g
```

PR #1253, opened 2023-03-28
Skip upstream sync on versions outside K8S_GO_MAP

addyess:
* skip upstream branches when no GO version is yet specified for the k8s version
* remove deprecated called to `semver.compare` replaced with `VersionInfo.compare`

### DIFF
--- a/cilib/service/snap.py
+++ b/cilib/service/snap.py
@@ -59,19 +59,24 @@ class SnapService(DebugMixin):
                 snapcraft_fn = src_path / "snapcraft.yaml"
                 snapcraft_fn_tpl = src_path / "snapcraft.yaml.in"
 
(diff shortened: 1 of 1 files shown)


cilib/service/snap.py
```diff
@@ -59,19 +59,24 @@ def sync_from_upstream(self):
...
+                    self.log(
+                        f"Skipping {k8s_version} because {k8s_major_minor} isn't in K8S_GO_MAP"
+                    )
+                    continue
 
                 snapcraft_yml_context = {
                     "snap_version": branch.lstrip("v"),
                     "patches": [],
-                    "go_version": enums.K8S_GO_MAP[
-                        f"{k8s_major_minor.major}.{k8s_major_minor.minor}"
```
kwmonroe, line 68: smh at whomever let this through in the first place

kwmonroe (approved):

PR #1260, opened 2023-04-05
include volcano image resources when releasing charm

addyess:
I've pushed some initial resources to charmhub for these charms, so builds will succeed, but any new charm changes should trigger resource pushes


mateoflorido (approved):
LGTM!

PR #1262, opened 2023-04-05
Cloud Provider Integration Testing

addyess:
Rectifies the following situations:
* juju-crashdump missed called arguments during test failures
* apt-install in a container occurs too quickly after container startup
* a `cloud-provider` or `storage-provider` doesn't include a charm to install/remove
* a `cloud-provider` and `storage-provider` are in the same charm (by sequentially deploying)
*  the original overlay creates a situation where a missing cloud provider would result in a **waiting/idle** `kubernetes-control-plane` unit

### DIFF
--- a/ci.bash
+++ b/ci.bash
@@ -167,12 +167,17 @@ function test::report
 function test::capture
 {
     if which juju-crashdump; then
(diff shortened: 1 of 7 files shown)


kwmonroe (approved):
LGTM

PR #1263, opened 2023-04-10
add filter_by_tag to sync jobs

kwmonroe:
Ensure our sync jobs process both k8s and k8s-operator tagged sources by default.

These jobs would previously take multiple `--filter-by-tag` click options, but we don't use that pattern in current jobs. Instead, other jobs rely on `FILTER_BY_TAG` being `.split(',')` and checked against the list of tags defined in our `includes/charm-support-matrix.inc`.

---

- [x] Needs `jjb` after merge: `sync.py`

### DIFF
--- a/jobs/sync-upstream.yaml
+++ b/jobs/sync-upstream.yaml
@@ -199,6 +199,9 @@
       - string:
           name: BUNDLE_REV
           description: bundle revision to tag stable branches with
+      - string:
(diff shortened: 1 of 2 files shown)


addyess (approved):
nice... this is similar to [build-charms](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/build-charms/charms.py#L988-L992)

PR #1265, opened 2023-04-11
build volcano bundle

addyess:
using the `bundle` shell scripts in volcano, build and publish its bundle:

Requires the following in `main` and `release_1.27` branch
* https://github.com/charmed-kubernetes/charm-volcano/pull/8


mateoflorido (approved):
LGTM!

PR #1271, opened 2023-04-14
1.27 stable work: adjust enums, start building 1.28

kwmonroe:
Set enums for 1.27 being the new stable release and enable 1.28/edge.

---

- [x] Needs `jjb` after merge: build-snaps.yaml


addyess (approved):
Rock on! 🎸

PR #1275, opened 2023-04-21
set min version of support to kubernetes dashboard to 1.28

addyess:
Charmed-Kubernetes will begin officially supporting kubernetes-dashboard in 1.28


mateoflorido (approved):
LGTM ✨

PR #1278, opened 2023-04-25
disable the validate-ck-390x job

addyess:
Disables the validate-ck-390x job

---

- [ x ] Needs `jjb` after merge

### DIFF
--- a/jobs/validate.yaml
+++ b/jobs/validate.yaml
@@ -305,6 +305,7 @@
     node: runner-validate
     description: |
       Validates CK edge for n-1 releases on s390x localhost.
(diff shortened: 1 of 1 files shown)


jobs/validate.yaml
```diff
@@ -305,6 +305,7 @@
     node: runner-validate
     description: |
       Validates CK edge for n-1 releases on s390x localhost.
+    disabled: true
```
kwmonroe, line 308: ha!  never would have guessed this key ;)

kwmonroe (approved):

PR #1280, opened 2023-04-26
Ceph Test refactor

mateoflorido:
# Changes 
- Changed the test logging to use `log` instead of `click`.
- Added a new Debug Method to gather logs in case the Ceph test fails. This should make it easier to diagnose the Ceph PVC, pods and provisioners.
- Changed the container pod images from `ubuntu` (~26MB) to `busybox` (~2.5MB). This should reduce the size when pulling from rocks.c.c.


jobs/integration/utils.py
```diff
@@ -418,32 +420,31 @@ async def validate_storage_class(model, sc_name, test_name):
...
+      command: ["/bin/cat"]
+      args: ["/mnt/data/hello.txt"]
+      volumeMounts:
+      - name: shared-data
+        mountPath: /mnt/data
   volumes:
   - name: shared-data
     persistentVolumeClaim:
       claimName: {sc_name}-pvc
-      readOnly: false
```
addyess, line 433: this probably could be `readOnly: true` don't you think if it's true this pod won't write to the volume

addyess (approved):
thanks @mateoflorido for the extra debug and use of a smaller image

PR #1281, opened 2023-04-26
test_audit_* tests should cleanup on failures, and simplify test reso…

addyess:
replacing the `run_until_success` calls that never timeout with simpler `kubectl` calls that will fail the test on failed `kubectl` is a safer way to ensure the suite of tests do not get stuck on an unresponsive API. 

testing of snap `1.28.0-alpha0` results in `test_audit_webhook` [failure](https://paste.ubuntu.com/p/64S3B97Qts/)

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -40,6 +40,8 @@ from .utils import (
     do_series_upgrade,
     finish_series_upgrade,
     kubectl,
+    kubectl_apply,
+    kubectl_delete,
     juju_run,
     juju_run_action,
     get_ipv6_addr,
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -1151,33 +1153,6 @@ async def all_certs_in_place():
...
-    raw = await run_until_success(unit, "du -bs /root/cdk/audit")
-    size_in_bytes = int(raw.split()[0])
-    click.echo("Audit log size in bytes: %d" % size_in_bytes)
-    max_size_in_bytes = 1000 * 1000 * 1000 * 1.01  # 1 GB, plus some tolerance
-    assert size_in_bytes <= max_size_in_bytes
-
-    # Clean up
-    await reset_audit_config(app, tools)
-
-
```
addyess, line 1180: moved down the file to join other `test_audit-*` methods

PR #1283, opened 2023-05-02
Debug ceph storage by defining K8sResources to debug and logging each

addyess:
This is a refactor of `_debug_ceph_storage` so i can reuse a basic structure to debug dns tests

### DIFF
--- a/ci.bash
+++ b/ci.bash
@@ -180,7 +180,7 @@ function test::capture
             -j snap.cdk-addons* \
             -m "$JUJU_CONTROLLER:$JUJU_MODEL"
     fi
-    tar -cvzf artifacts.tar.gz ci.log _out meta juju-crashdump* report.* failures* || true
+    tar -cvzf artifacts.tar.gz ci.log _out meta juju-crashdump* report.* failures* logs/ || true
     /usr/local/bin/columbo -r columbo.yaml -o "_out" "artifacts.tar.gz" || true
     python bin/s3 cp "columbo-report.json" columbo-report.json || true
 
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -12,6 +12,7 @@ import shlex
 import uuid
 import yaml
 
+from contextlib import contextmanager
 from datetime import datetime
 from juju.model import Model
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


ci.bash
```diff
@@ -180,7 +180,7 @@ function test::capture
             -j snap.cdk-addons* \
             -m "$JUJU_CONTROLLER:$JUJU_MODEL"
     fi
-    tar -cvzf artifacts.tar.gz ci.log _out meta juju-crashdump* report.* failures* || true
+    tar -cvzf artifacts.tar.gz ci.log _out meta juju-crashdump* report.* failures* logs/ || true
```
addyess, line 183: including the `logs/` directory in the zipped artifact

jobs/integration/conftest.py
```diff
@@ -470,14 +471,30 @@ def skip_by_model(request, model):
...
+        if "mode" not in kwargs:
+            # defaults to writing log files if mode not specified
+            kwargs["mode"] = "w"
+        with path.open(**kwargs) as fp:
+            log(f"Logging to {path}... ")
+            yield fp
+
+    yield _open
+
+
```
addyess, line 497: this fixture yields a method which can be used like python's `open` in a with context to write log files in a directory based on the name of the current running test.

jobs/integration/utils.py
```diff
@@ -464,12 +468,13 @@ async def validate_storage_class(model, sc_name, test_name):
...
-            pods=[f"{sc_name}-write-test", f"{sc_name}-read-test"],
-        )
-
+    except:
+        # bare except intentional to debug any failure
+        if debug_open:
+            await _debug_storage_class(
+                debug_open, test_name, sc_name, provisioner, model
+            )
+        raise
```
addyess, line 477: if we catch ANY error -- let's debug that error but don't continue the testing -- whichever exception needs to be re-raised so the test doesn't accidentally PASS

jobs/integration/utils.py
```diff
@@ -496,37 +501,40 @@ async def validate_storage_class(model, sc_name, test_name):
         )
 
 
-async def _debug_ceph_storage(control_plane, pods):
-    log.error("Gathering Ceph logs from Cluster")
+async def _debug_storage_class(debug_open, test_name, sc_name, provisioner, model):
+    class _Call:
+        def __init__(self, *args, **kwds):
+            self.args, self.kwds = args, kwds
```
addyess, line 507: class to hold variable args and kwargs

jobs/integration/utils.py
```diff
@@ -496,37 +501,40 @@ async def validate_storage_class(model, sc_name, test_name):
...
+    provisioner_pods = []
+    if provisioner:
+        result = await kubectl(
+            model,
+            "get",
+            "pods",
+            A=True,
+            l=f"app={provisioner}",
+            o=r"""jsonpath='{range .items[*]}{@..metadata.namespace}{" "}{@..metadata.name}{"\n"}{end}'""",
+        )
```
addyess, line 521: this `jsonpath` formats the pods as `<pod-namespace> <pod-name` per line of output

using that, we can parse the output lines and look up the logs of the provisioner further down

mateoflorido (approved):
LGTM ✨

PR #1297, opened 2023-05-09
Replace install-mode with "disable" for kubelet-eks

toabctl:
When the kubelet-eks service is automatically started after installation, it will fail because the service is usually not correctly configured.
This results in lots of logs with error messages and also with restart rate-limit problems with systemd.
So "disable" the service by default and enable it when ready.

---

- [ ] Needs `jjb` after merge


addyess:
I don't believe the jjb is necessary, just a merge should do

Cynerva (approved):

PR #1312, opened 2023-05-16
Sync `prometheus-manual` interface from upstream source

addyess:
charms like `etcd` use this interface to sync with `prometheus2` charm, but were missing some key commits in `charmed-kubernetes` org


kwmonroe (commented):
i mean, can't hurt.

PR #1319, opened 2023-05-18
Ensure install-mode enable for kubelet-eks

rpocase:
This seems to have [introduced regressions](https://bugs.launchpad.net/cloud-images/+bug/2020072) in live clusters and new images. We'll need to dig into this further before going this route.

---

- [ ] Needs `jjb` after merge

- For integration tests, please make changes in `jobs/integration`
- For validation envs, `jobs/validate`
- For MicroK8s,`jobs/microk8s`
- For charm/bundle builds, `jobs/build-charms`


rthill91 (approved):

JessicaJang (approved):

Cynerva (approved):
LGTM. Do you need us to rebuild snaps with this change, or are you able to roll back the kubelet-eks revisions in the snap store?

rpocase:
@Cynerva Just due to personal unfamiliarity, I think rebuilding would be safer to get all the channels updated together. I believe Tom has handled this in the past and I worry I'd break revert to the wrong version :).

PR #1320, opened 2023-05-19
Set kubelet-eks service install-mode to disable

rpocase:
After further digging, the upstream issues seem to be caused by the refresh cycle, not this change specifically. So, reverting the revert. Please trigger rebuilds one more time once this is merged, thanks!

---

- [ ] Needs `jjb` after merge

- For integration tests, please make changes in `jobs/integration`
- For validation envs, `jobs/validate`
- For MicroK8s,`jobs/microk8s`
- For charm/bundle builds, `jobs/build-charms`

### DIFF
--- a/jobs/build-snaps/build-release-eks-snaps.groovy
+++ b/jobs/build-snaps/build-release-eks-snaps.groovy
@@ -56,7 +56,7 @@ pipeline {
                         cd \${EKS_SNAP}
                         sed -i -e "s/^name: \${snap}/name: \${EKS_SNAP}/" \
                                -e "s/^base: .*/base: ${EKS_BASE}/" \
-                               -e "s/install-mode: .*/install-mode: enable/" snapcraft.yaml
+                               -e "s/install-mode: .*/install-mode: disable/" snapcraft.yaml
 
                         # if we don't have any base defined at this point, add one
                         grep -q "^base: " snapcraft.yaml || echo "base: ${EKS_BASE}" >> snapcraft.yaml


Cynerva:
If there's a problem with the refresh cycle, and we rebuild now, is that going to cause another wave of refreshes and hence more problems? Or has that been addressed separately?

rpocase:
That's a relevant concern. We are discussing using `snap hold` for future AMIs, but we can't do anything about production clusters today.  We should still stage this change for the next patch cycle.

If we revert to a previous version via the store would that trigger a refresh? I'll have to double check if account has the right access, but I'm definitely +1 to that route if it avoids the possible concern.

Cynerva:
I think reverting in the store would still lead to VMs refreshing. It *might* prevent the break in VMs that are refreshing to a revision that they previously had installed, since the `/var/snap/kubelet-eks/$REV/args` file may still exist on the filesystem when the refresh occurs. But I'm not sure.

I think the refresh issue can be fixed in the snap. Along with setting `install-mode: disable`, I want to try the following:
1. In the [wrapper script](https://git.launchpad.net/snap-kubelet/tree/wrappers/run-with-config-args) that reads the args file, if the args file doesn't exist, exit without starting kubelet. This will ensure kubelet doesn't start in standalone mode.
2. In the configure hook (generated [here](https://git.launchpad.net/snap-kubelet/tree/shared/generate-configure-hook)), write the args file atomically (write to args.tmp then move it) and after writing it, restart the kubelet service.
3. In snapcraft.yaml, set `restart-delay=4s` (documented [here](https://snapcraft.io/docs/services-and-daemons)) to reduce the frequency of restarts and hopefully prevent hitting the systemd start rate limit (default 5 restarts in 10 seconds)

Would you be open to me testing these today and trying to get them in prior to a rebuild? I want to get this fixed for you quickly, but I also *really* don't want to break any more live nodes if we can avoid it.

Cynerva:
I'm gonna go ahead and merge this now, but not starting builds yet without further feedback. See my previous comment and https://code.launchpad.net/~k8s-jenkaas-admins/snap-kubelet/+git/snap-kubelet/+merge/443294 which will hopefully fix the failures occurring with snap refreshes.

rpocase:
Thanks for the follow up! Sorry I missed your reply earlier - don't have
this org setup with a proper email redirect yet.

I'm very happy with your proposed solution and the linked MP looks good to
me. It's in line with what we already scopes out and I'm happy to drop it
from our backlog :).

PR #1327, opened 2023-05-24
Skip `test_service_cidr_expansion` when testing kubeovn and multus

addyess:
[LP#2020704](https://bugs.launchpad.net/charm-multus/+bug/2020704) Multus DaemonSet should be restarted during cidr expansion

* since it isn't restarted, lets skip this test until its fixed


Cynerva (approved):
:+1:

PR #1342, opened 2023-06-07
Back out 1.28/candidate from from k8s track list

Cynerva:
As I understand it, the inclusion of `1.28/candidate` in the k8s track list  is causing build-charms, when targeted to channel `candidate`, to publish to `1.28/candidate` instead of `1.27/candidate`. This is creating a lot of confusion within our team for the release process.

Is it something we can just back out now and re-add later?


Cynerva:
I think we might also need to revert https://github.com/charmed-kubernetes/jenkins/pull/1316 to get the bugfix release doc back on track too. Or generate a new image that targets `candidate` like it did before

kwmonroe:
> get the bugfix release doc back on track too. Or grab a new screenshot that targets `candidate` like it did before

I have this along with a few other bugfix doc fixes that i'll drop in for [LP:2023508](https://bugs.launchpad.net/charmed-kubernetes-testing/+bug/2023508).

kwmonroe (approved):
This cleared up some confusing test runs, thanks!

PR #1345, opened 2023-06-20
Handle fail cases where a k8s resource used in ceph tests wasn't created

addyess:
I found situations where the `finally` case of the `test_ceph` raised assertions that it couldn't delete a `read-test` pod because it didn't exist.  Unfortunately there is a test failure, but it's hidden because of this failure in the `finally` section. 

The `finally` clause should ATTEMPT to delete what's there -- but ignore resources that aren't there.

### DIFF
--- a/jobs/integration/utils.py
+++ b/jobs/integration/utils.py
@@ -469,16 +469,11 @@ spec:
         raise
     finally:
         log.info(f"{test_name}: {sc_name} cleanup")
-        pods = "{0}-read-test {0}-write-test".format(sc_name)
+        kubectl_delete = "/snap/bin/kubectl --kubeconfig /root/.kube/config delete --ignore-not-found=true"
+        pods = f"{sc_name}-read-test {sc_name}-write-test"
         pvcs = f"{sc_name}-pvc"
-        pod_deleted = await juju_run(
-            control_plane,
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -1462,25 +1462,27 @@ def _find_relation(*specs):
...
+    @pytest.mark.parametrize("storage_class", ["ceph-xfs", "ceph-ext4", "cephfs"])
+    async def test_storage_class(self, model, log_open, storage_class):
+        # create pod that writes to a pv from ceph
+        kwds = dict(debug_open=log_open)
+        if storage_class == "cephfs":
+            kwds["provisioner"] = "cephfs-provisioner"
+        else:
+            kwds["provisioner"] = "csi-rbdplugin-provisioner"
+
+        await validate_storage_class(model, storage_class, "Ceph", **kwds)
```
addyess, line 1485: split the previous `test_ceph` into what amounts to 4 tests -- one to see if the plugins are installed correctly, and 3 for each of the storage classes associated with these charms

the `ceph_apps` pytest fixture is changed from a function-scoped fixture that cleans up the apps on completion of the test function, to class-scoped which cleans up when all the tests are completed associated with the `TestCeph` class.


Finally -- move the check to skip the tests into the `ceph_apps` fixture rather than use `skip_if_apps` marker

PR #1361, opened 2023-07-11
remove nightly builds of metallb-speaker/controller charms

addyess:
metallb is now a single charm, rather than 2 independent charms.

Remove the nightly build tag (it can still be built using the `metallb` tag if targeting an older branch (like 1.27 or prior)

### DIFF
--- a/jobs/includes/charm-support-matrix.inc
+++ b/jobs/includes/charm-support-matrix.inc
@@ -439,7 +439,6 @@
     summary: Controller charm for the metallb loadbalancer
     tags:
       - docs-extra
-      - metallb
       - metallb-controller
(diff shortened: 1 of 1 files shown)


jobs/includes/charm-support-matrix.inc
```diff
@@ -453,7 +452,6 @@
     subdir: charms/metallb-speaker
     summary: Speaker charm for the metallb loadbalancer
     tags:
-      - k8s
       - docs-extra
       - metallb
```
mateoflorido, line 456:
```suggestion
```

mateoflorido (approved):
LGTM ✨

PR #1367, opened 2023-07-13
cephfs validation should collect logs from cephfs provisioner

addyess:
in order to appropriately collect logs from cephfs failures, the correct cephfs provisioner logs should be acquired

These changes also allow the ceph tests to run against models where ceph-fs, ceph-mon, and ceph-osd are currently installed and idle

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -1045,14 +1045,15 @@ async def test_service_cidr_expansion(model):
     await app.set_config(new_config)
     await wait_for_process(model, service_cluster_ip_range)
 
-    cmd = "/snap/bin/kubectl --kubeconfig /root/.kube/config get service kubernetes"
-    control_plane = model.applications["kubernetes-control-plane"].units[0]
-    output = await juju_run(control_plane, cmd, check=False)
+    output = await kubectl(model, "get service kubernetes")
     assert output.status == "completed"
 
     # Check if k8s service ip is changed as per new service cidr
     raw_output = output.stdout
     assert new_service_ip_str in raw_output
+
+    # Wait for the model to be stable
+    await model.wait_for_idle(status="active", timeout=10 * 60)
 
 
 async def test_sans(model):
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -2527,6 +2527,9 @@ async def ceph_apps(model, tools):
 
     all_apps = ["ceph-mon", "ceph-osd", "ceph-fs"]
     if any(a in model.applications for a in all_apps):
+        if not tools.use_existing_ceph_apps:
+            pytest.skip("Skipped since ceph apps are already installed")
+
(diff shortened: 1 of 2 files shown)


jobs/integration/validation.py
```diff
@@ -1045,15 +1045,16 @@ async def test_service_cidr_expansion(model):
     await app.set_config(new_config)
     await wait_for_process(model, service_cluster_ip_range)
 
-    cmd = "/snap/bin/kubectl --kubeconfig /root/.kube/config get service kubernetes"
-    control_plane = model.applications["kubernetes-control-plane"].units[0]
-    output = await juju_run(control_plane, cmd, check=False)
+    output = await kubectl(model, "get service kubernetes")
```
kwmonroe, line 1048: :heart:

jobs/integration/validation.py
```diff
@@ -2524,8 +2525,13 @@ async def ceph_apps(model, tools):
     elif series_idx > SERIES_ORDER.index("jammy"):
         pytest.fail("ceph_charm_channel is undefined past jammy")
 
-    if any(a in model.applications for a in ["ceph-mon", "ceph-osd"]):
-        pytest.skip("Skipped since ceph apps are already installed")
+    all_apps = ["ceph-mon", "ceph-osd", "ceph-fs"]
+    if any(a in model.applications for a in all_apps):
+        # allow currently deployed ceph apps to run tests
```
kwmonroe, line 2530: hmm, i'm worried about this.  i have a strong remembrance of skipping pre-existing ceph deployment testing.  as in, if `validation.py` didn't deploy it, we don't want to test it since that could affect other products relying on this model.

Perhaps it's ok as we're yielding / returning here and won't fall into the ceph removal below, but i'd like a +2 here in case we need to be careful about running [validate_storage_class](https://github.com/charmed-kubernetes/jenkins/blob/f1e382ac3862c386a2905950ebaa6b15e3f23b76/jobs/integration/utils.py#L366) on a preexisting ceph.

@Cynerva, wadda you think?

kwmonroe (commented):
Tentative +1 from me, but i'd like a second opinion around testing on pre-deployed ceph units.

jobs/integration/validation.py
addyess: i can remove it -- i had it in there to be able to run the tests over and over without the ceph setup/teardown

addyess: or I could add a test option like `--use-existing-ceph-apps` defaulted to `False`

I can run my tests -- but the normal runs wouldn't be touched

Cynerva: This seems fine to me. If Ceph is in the same model as CK then I think it's fair to assume that they're supposed to be integrated. And if they are, then when I run validation, I would expect that integration to be tested.

addyess: ok, i'm gonna take that as a LGTM...

PR #1376, opened 2023-07-21
Adjustments to enable building on launchpad

addyess:
In order to build charms targeting a certain architecture, the intent is to leverage a build farm of launchpad builders using  the `launchpadlib` library within the charm builder job

By changing only the `charm-support-matrix.inc` to include the `builder: 'launchpad'` option within a specific charm, the charm will be built on launchpad rather than on the jenkins runner.  

It is necessary only for the charm to be built using `charmcraft.yaml` (no reactive charms are supported there) and each `builds-on` entry in that `charmcraft yaml` will recieve a separate builder.

Incidentally, this allows for a single charm build invocation to also yield multiple charm build outputs -- one for each `builds-on` entry.  Much of the rearchitecture in this PR is around handling a single charm build invocation creating multiple charm files (eg. one charm file per architecture)

In that event -- these changes support each architecture being uploaded, associated with its resources, and released to the same charmhub channel -- for the various architectures|bases it supports.

These changes also facilitate having a different architecture's OCI image attached to a charm so that juju can download the appropriate binary arch and binary oci-image for that charm deployment. 

- [ ] update build-charms / promote charms job jjb completed

### DIFF
--- a/jobs/build-charms.yaml
+++ b/jobs/build-charms.yaml
@@ -35,7 +35,7 @@
           COMMAND: |
             #!/bin/bash
             set -eux
-            tox -e py38 -- python jobs/build-charms/charms.py promote \
+            tox -e py38 -- python jobs/build-charms/main.py promote \
                 --from-channel $FROM_CHANNEL \
                 --to-channel $TO_CHANNEL \
                 --charm-list $CHARM_LIST \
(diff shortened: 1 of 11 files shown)


mateoflorido (approved):
LGTM 💯

PR #1377, opened 2023-07-24
Create a pytest marker to skip if a bug was unresolved, and skip cephfs test

addyess:


jobs/integration/validation.py
```diff
@@ -1474,15 +1474,20 @@ async def test_plugins_installed(self, model):
             timeout_msg="CSI pods not ready!",
         )
 
-    @pytest.mark.parametrize("storage_class", ["ceph-xfs", "ceph-ext4", "cephfs"])
+    @pytest.mark.parametrize(
+        "storage_class",
+        [
+            "ceph-xfs",
+            "ceph-ext4",
+            pytest.param("cephfs", marks=pytest.mark.xfail_if_open_bugs(2028387)),
```
addyess, line 1482: xfail the `cephfs` test only

jobs/integration/conftest.py
```diff
@@ -601,6 +602,21 @@ async def k8s_version(model):
...
 
 
+@pytest.fixture(autouse=True)
+def xfail_if_open_bugs(request):
+    xfail_marker = request.node.get_closest_marker("xfail_if_open_bugs")
+    if not xfail_marker:
+        return
+    bugs = xfail_marker.args
+    lp = LPClient()
+    lp.login()
```
Cynerva, line 612: Hmm, I'm not crazy about the dependence on a launchpad connection to run this test suite. Two issues that come to mind:
1. If launchpad is down, this will error. (or hang?)
2. If running tests locally, we now have to have local credentials and the LPCREDS environment variable set, otherwise this will error.

These could impact SQA if they are running marked tests in their SKUs.

The good news is that only the marked tests will be impacted. Other tests should still work fine without any launchpad credentials or connectivity.

addyess: so, the likelihood of not having credentials is probably higher than launchpad itself going down. 

I'll update the Launchpad client code to try to use anonymous login -- bug status should still be visible.  I guess if launchpad itself is down -- these tests should just "run as normal" -- if they fail, they fail and we can't mark as `xfail`

Cynerva: Seems like a reasonable compromise. Sounds good to me, thanks :+1:

Cynerva (approved):

PR #1393, opened 2023-08-02
Map the snap base core to the eks version number

addyess:
Provide an EKS snap base override as well as map snap bases to specific kubernetes versions

### DIFF
--- a/jobs/build-snaps/build-release-eks-snaps.groovy
+++ b/jobs/build-snaps/build-release-eks-snaps.groovy
@@ -3,6 +3,23 @@
 def kube_version = params.k8s_tag
 def kube_ersion = kube_version.substring(1)
 def lxc_name = env.JOB_NAME+"-"+env.BUILD_NUMBER
+def _find_eks_base(version, override){
+    if (override.length())
+        return override
+    def EKS_BASE_DEFAULT = "core20"  // If not in the BASES map, use this as a default
+    def EKS_BASES = [
+        core22: {float i -> 1.28 <= i},             // 1.28 and beyond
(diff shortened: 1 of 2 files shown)


jobs/build-snaps/build-release-eks-snaps.groovy
```diff
@@ -3,6 +3,23 @@
...
+        core18: {float i -> 1.20 <= i && i < 1.24}, // 1.20 <= v < 1.24
+    ]
+    // assumes the version looks like r'v\d.\d+.\d+'
+    def eks_base_key = Float.parseFloat(version.substring(1, version.lastIndexOf('.')))
+    // find the first matching predicate
+    def eks_found_base = EKS_BASES.find{it -> it.value(eks_base_key)}
+    return eks_found_base ? eks_found_base.key : EKS_BASE_DEFAULT
+}
+def eks_base_override = params.eks_base_override
+def EKS_BASE = _find_eks_base(kube_version, eks_base_override)
```
Cynerva, line 22: I think this might need to be in the environment for the shell scripts? I'll take a closer look in a bit

addyess: it very well may... thanks for peaking at this

Cynerva: Ah nevermind, this looks fine actually. Seems like it's put in the shell scripts via Groovy string formatting, the same way something like the global `kube_version` is. :+1:

addyess:
per @toabctl 

>  core20 is definitely used since 1.22 (maybe even before that). But it
    doesn't matter because 1.22 is already EOL. So having core20 for
    everything < 1.28 and core22 for everything >= 1.28 should be good.

toabctl (approved):

PR #1398, opened 2023-08-11
Add lightkube to the list of forks kept in sync by cdkbot

addyess:
The sync job synchronizes a number of "Layers and Charms" from their upstream source, to their downstream source.  By adding lightkube to the list of "layer" -- then this repo will be kept in sync with its upstream source. 

It doesn't need stable branches, tagging, or to support renaming -- and it isn't necessary to pull locally for the reactive build cache.

### DIFF
--- a/jobs/build-charms/builder_local.py
+++ b/jobs/build-charms/builder_local.py
@@ -540,7 +540,8 @@ class BuildEnv:
         layers_to_pull = [
             layer_name
             for layer in self.layers
-            if (layer_name := next(iter(layer))) != "layer:index"
+            for layer_name, layer_ops in layer.items()
+            if layer_ops.get("build_cache") != False
         ]
         pool = ThreadPool()
         results = pool.map(self.download, layers_to_pull)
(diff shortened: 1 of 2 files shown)


jobs/build-charms/builder_local.py
```diff
@@ -540,7 +540,8 @@ def pull_layers(self):
         layers_to_pull = [
             layer_name
             for layer in self.layers
-            if (layer_name := next(iter(layer))) != "layer:index"
+            for layer_name, layer_ops in layer.items()
+            if layer_ops.get("build_cache") is not False
```
kwmonroe, line 544: dangerously clever here, i love it!

kwmonroe (approved):
This lgtm.  Please confirm that cdkbot can indeed push to the canonical org (vs c-k) before merging.

addyess:
> This lgtm. Please confirm that cdkbot can indeed push to the canonical org (vs c-k) before merging.

Yep, it can.  I tested it this morning

jobs/build-charms/builder_local.py
addyess: If you think this is dangerous -- i'll be happy to do something more clear

kwmonroe: nah, all good -- i like it using this `build_cache` opt vs matching the layer name.

PR #1404, opened 2023-08-16
Improve exception tracing when setting up a k8s-cloud

addyess:
We continue to see errors in the validation tests where `k8s_cloud` fixture is involved, but the only error coming back has to do with a failed async.  This is because the fixture is trying too hard to catch exceptions -- when it should really bubble those out

There's also a bit of flakiness in many tests to get the `kubeconfig` from the control-plane unit.  lets setup one fixture that can do that for every test.  If we don't like it being module scoped -- that's fine we can remove that.

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -19,15 +19,14 @@ from juju.model import Model
 from pathlib import Path
 from py.xml import html
 from tempfile import NamedTemporaryFile
-from traceback import format_exc
 from .utils import (
     asyncify,
     upgrade_charms,
     upgrade_snaps,
     arch,
     log_snap_versions,
-    scp_from,
     juju_run,
+    juju_run_action,
 )
 
 from .logger import log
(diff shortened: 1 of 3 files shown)


jobs/integration/conftest.py
```diff
@@ -301,53 +301,42 @@ async def model(request, tools):
...
-            kcp_unit,
-            "config",
-            f.name,
+async def k8s_cloud(kubeconfig, tools):
+    clouds = await tools.run(
+        "juju", "clouds", "--format", "yaml", "-c", tools.controller_name
+    )
+    if tools.k8s_cloud in yaml.safe_load(clouds[0]):
+        yield tools.k8s_cloud
+        return
```
addyess, line 310: first -- let's see if the cloud we want already exists -- to prevent a failure when adding it a second time

jobs/integration/conftest.py
```diff
@@ -301,53 +301,42 @@ async def model(request, tools):
...
+        await tools.run(
+            "juju",
+            "add-k8s",
+            "--skip-storage",
+            "-c",
             tools.controller_name,
-            tools.connection,
-            proxy=tools.juju_ssh_proxy,
+            tools.k8s_cloud,
         )
```
addyess, line 322: next, let's `add-k8s`.  The `kubeconfig` fixture will handle adding the correct `kubeconfig` file into the `os.environment`

jobs/integration/conftest.py
```diff
@@ -301,53 +300,42 @@ async def model(request, tools):
...
-        try:
-            click.echo("Adding k8s cloud")
-            os.environ["KUBECONFIG"] = f.name
+        _created = True
+        yield tools.k8s_cloud
+    finally:
+        if _created:
+            click.echo("Removing k8s cloud")
```
addyess, line 326: the remove portion is only run if `_created` was set

jobs/integration/conftest.py
```diff
@@ -356,35 +344,36 @@ async def k8s_model(k8s_cloud, tools):
...
+        await _model_created.connect(tools.k8s_connection)
+        yield _model_created
     finally:
-        if not k8s_model:
-            return
-        await tools.run("juju-crashdump", "-a", "config", "-m", tools.k8s_connection)
-        click.echo("Cleaning up k8s model")
-        try:
-            for relation in k8s_model.relations:
+        if _model_created:
```
addyess, line 357: we're careful here to not `return` in the finally, so exceptions can freely bubble out

jobs/integration/conftest.py
```diff
@@ -356,35 +344,36 @@ async def k8s_model(k8s_cloud, tools):
...
-        yield k8s_model
+        _model_created = Model()
+        await _model_created.connect(tools.k8s_connection)
+        yield _model_created
     finally:
-        if not k8s_model:
-            return
-        await tools.run("juju-crashdump", "-a", "config", "-m", tools.k8s_connection)
-        click.echo("Cleaning up k8s model")
-        try:
```
addyess, line 373: also -- stop trying to catch errors in the finally -- just let the exceptions flow so we can find the issues

kwmonroe (approved):

PR #1413, opened 2023-08-28
Ensure that lp builders are insync with upstream sources

addyess:
Ensure the commit hash from the launchpad git sources are insync with the upstream sources before building the charm on a launchpad builder.


Cynerva (approved):
Oh nice :+1: Thanks for this, LGTM

PR #1414, opened 2023-08-29
clean up the deployment before destroying the namespace

addyess:
Addressing test timeouts during the teardown of the autoscaler tests.  

[LP#2033414](https://bugs.launchpad.net/charm-kubernetes-autoscaler/+bug/2033414)

### DIFF
--- a/jobs/integration/test_autoscaler.py
+++ b/jobs/integration/test_autoscaler.py
@@ -29,6 +29,7 @@ async def deployment(kubectl, namespace):
     kubectl.create.namespace(namespace)
     kubectl.create(f=path_to_deployment, namespace=namespace)
     yield path_to_deployment
+    kubectl.delete(f=path_to_deployment, namespace=namespace, grace_period=5*60)
     kubectl.delete.namespace(namespace)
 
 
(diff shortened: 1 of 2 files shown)


jobs/integration/test_autoscaler.py
```diff
@@ -29,6 +29,7 @@ async def deployment(kubectl, namespace):
     kubectl.create.namespace(namespace)
     kubectl.create(f=path_to_deployment, namespace=namespace)
     yield path_to_deployment
+    kubectl.delete(f=path_to_deployment, namespace=namespace, grace_period=5 * 60)
```
kwmonroe, line 32: why wait 5m vs the default 30s here?

kwmonroe (approved):
+1 as is, as well as +1 for other grace period timeouts if we don't *need* to wait the full 5m.

jobs/integration/test_autoscaler.py
addyess: maybe this is the wrong timeout value.  Currently the following command (the one to delete the namespace) never completes.  It just sits there forever.  I was hoping by adding a grace-period, it would quit after 5m.  Maybe i'm reading this doc wrong

PR #1416, opened 2023-08-30
provide support for arch specific file resources

addyess:
If a multi-arch charm has a file resource which must be different based on the charm's arch (for instance, it contains a binary image) then the build tool chain needs to be able to release the correct arch based resource along with that arch specific charm artifact.

* This allows charms to assemble an arch specific file resource
* This allows a channel-range to be applied for a specific charm resource 
    * ex) in the 1.28 track this charm grew a need for a foo resource
    * This lets bug fix releases continue to release the charm without that resource

### DIFF
--- a/cilib/version.py
+++ b/cilib/version.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional, Union
+from typing import Optional, Mapping, Union
 import semver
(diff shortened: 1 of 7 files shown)


mateoflorido (approved):
This is great, LGTM! Thanks @addyess 🎆

PR #1441, opened 2023-10-31
Prevent o7k CCM from syncing from a repo we're closing

addyess:
Stop syncing from the canonical org since it will be closed when [160639](https://portal.admin.canonical.com/C160639) is completed


mateoflorido (approved):
LGTM

PR #1444, opened 2023-11-03
Speed up `cidr_expansion` by firing the update-status hooks more often

addyess:
Adding a context loop to change the update-status hook retries rather than the default 5m

* drive-by upgrade of juju
* drive-by fix deprecation warning


mateoflorido (approved):
Great, LGTM!

PR #1457, opened 2023-12-05
Fix validation to work with ops charms

Cynerva:
Hopefully so we can start getting some validation runs against the ops charms.

I wasn't quite able to do a full run of this today - my cluster blew up when I ran `test_service_cidr_expansion`. Will try again tomorrow.

### DIFF
--- a/jobs/integration/validation.py
+++ b/jobs/integration/validation.py
@@ -2059,7 +2059,7 @@ async def test_sysctl(model, tools):
         await app.set_config({"sysctl": config["sysctl"]["value"]})
 
 
-async def test_cloud_node_labels(model, tools):
+async def test_cloud_node_labels(cloud, model, tools):
     unit = model.applications["kubernetes-control-plane"].units[0]
     cmd = "/snap/bin/kubectl --kubeconfig /root/.kube/config get no -o json"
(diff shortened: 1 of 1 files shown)


jobs/integration/validation.py
```diff
@@ -2071,24 +2059,16 @@ async def verify_sysctl(units, desired_values):
...
-    elif "azure-integrator" in model.applications:
-        assert label == "azure"
-    elif "gcp-integrator" in model.applications:
-        assert label == "gce"
-    elif "openstack-integrator" in model.applications:
-        assert label == "openstack"
-    elif "vsphere-integrator" in model.applications:
-        assert label == "vsphere"
+    if cloud in ["azure", "ec2", "gce", "openstack", "vsphere"]:
+        assert label == cloud
```
addyess, line 2071: hmm... clever!

addyess (approved):
Aight--- skipping some tests eh?

Cynerva:
Yeah. I don't really like it. The skipped tests are covering features and integrations that no longer exist. Some of those features might come back in a later release - that's the part I don't like.

I think we have to skip these tests for the initial 1.29 release though, given that our plan is to ship without those features, right?

kwmonroe (approved):

PR #1459, opened 2023-12-08
pick up golang candidate for 1.29.0-rc2 snap builds

kwmonroe:
Bump 1.29 enums to pick up candidate golang for our 1.29.0-rc2 snap builds; promote those builds to both 1.29 edge and beta channels.

This will need to be adjusted once go/1.21/stable is populated with 1.21.5.


addyess (approved):
Ack.  need a reminder to roll the GO_MAP backwards?

PR #1465, opened 2023-12-14
Check channel boundaries before checking out charm code

addyess:
`tigera-secure-ee` charm should not be built beyond the [1.28 release branch](https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/includes/charm-support-matrix.inc#L278-L291), and if queued for building during 1.29/beta, it should be skipped over due to its channel boundaries settings.  

This code will check the channel-boundaries and only build if there is a channel for a charm to be published to based on the channel boundaries


mateoflorido (approved):
LGTM

PR #1469, opened 2024-01-11
Include nvidia-operator charms to builds, excluding the nightly builds

addyess:
by not including the `k8s-operator` tag, these charms won't build nightly

But we can run a parameterized build to craft these charms one off. 

I've also requested tracks to be opened for these charms at 1.29.


kwmonroe (commented):
LGTM :)

PR #1482, opened 2024-03-06
new stable and pre-release defaults now that 1.29 is ga and 1.30 is open

kwmonroe:
set new stable and pre-release defaults now that 1.29 is ga and 1.30 tracks are open

---

- [x] Needs `jjb` after merge: affects `ci-master.yaml`; needs all the jjbs


jobs/ci-master.yaml
```diff
@@ -89,7 +89,7 @@
           name: juju_channel
           description: |-
             specify the juju channel to use
-          default: 3.1/stable
+          default: 3/stable
```
kwmonroe, line 92: any concerns here?  this will take our workers to 3.4 (as of today).

kwmonroe:
i just realized this thing's been hanging out in Draft forever.  Most of it was superseded by #1494, but we may want to move the default juju channel later -- not now during release week, but later.

addyess:
This is good stuff.  Thanks @kwmonroe .  I think we need the bandaid ripped off to flow with juju 3/stable

PR #1487, opened 2024-03-07
Update validation tests for google cloud

addyess:
Charmed-kubernetes starting with 1.29/edge supports `gcp-cloud-provider` integration. 

the `:gcp` relation with the worker and control-plane is available only for instance tagging, but isn't necessary for the operational storage or load-balancers


kwmonroe (approved):
Very cool to see the new `in-tree-until` for gcp-c-p.  nice job!

PR #1494, opened 2024-03-13
Update to testing with 1.30/edge channels

addyess:
Charms are available from 1.30/edge
Snaps are available from 1.30/edge
It's time to update the CI to test 1.30/edge

* [x] needs jjb


mateoflorido (approved):
LGTM!

PR #1496, opened 2024-03-19
Rather than use -y, use --no-prompt when destroying or killing juju controllers

addyess:
Since juju 3.2, `-y` option has been removed.  This leaves models sitting around in our CI un-destroyed after a test completes.


mateoflorido (approved):
LGTM

PR #1498, opened 2024-03-25
Reintroduce vault kv testing

addyess:
Enables vault-kv testing again,  this time actually ensuring etcd doesn't have the secret data in plain text


mateoflorido (approved):
LGTM

PR #1499, opened 2024-03-26
Replace release-microk8s job from groovy to shell script

addyess:
Removes microk8s-release groovy job and replaces it with a shell script based on the validate jobs

---

- [X] Needs `jjb` after merge

I believe the existing pipeline jobs will need to be deleted before these new matrix style jobs can be contributed


jobs/ci-master.yaml
```diff
@@ -5,7 +5,7 @@
       - git:
           url: https://github.com/charmed-kubernetes/jenkins.git
           branches:
-            - 'main'
+            - 'KU-486/microk8s-release-jobs-on-lxc'
```
addyess, line 8: Don't merge this!

berkayoz (commented):
Looks good overall, I've checked out some test runs and logs and release python scripts seem to kick in which seems good enough for the scope of this card. To note the scripts are failing, which will result in some more work.

addyess:
Thanks @berkayoz!  I appreciate the response

PR #1500, opened 2024-03-27
kubernetes-control-plane/worker no longer supports vsphere relation in 1.29 and beyond

addyess:
the core vsphere bundle no longer needs vsphere relations against kubernetes-worker/control-plane.  They haven't needed them for some time and provide no extra features other than supplied cloud credentials.

### DIFF
--- a/jobs/validate/integrator-spec
+++ b/jobs/validate/integrator-spec
@@ -63,9 +63,9 @@ EOF
       folder: k8s-crew-root
     num_units: 1
     trust: true
-relations:
-  - ['vsphere-integrator', 'kubernetes-control-plane:vsphere']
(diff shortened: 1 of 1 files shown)


kwmonroe (approved):
Lgtm. For other travelers, this one confused me for a minute -- how does v-c-p trigger the external-c-p for k-c-p?

It's here, from back when we moved cloud providers out-of-tree:

https://github.com/charmed-kubernetes/jenkins/blob/main/jobs/integration/templates/integrator-charm-data/vsphere/out-of-tree.yaml

There no need for a v-i integration anymore (in fact, it's an error to do so).

This correctly sets our vsphere env up for testing. Thank you Adam!

PR #1501, opened 2024-03-28
build microk8s 1.30

bschimke95:
I don't know exactly if we need a `jjb` but I guess so. Better safe than sorry :)

---

- [x] Needs `jjb` after merge


addyess:
@bschimke95 in this case you didn't change the jenkins job definitions (`/jobs/*.yaml`) files so there's no need for jjb after merge

I'll merge once the 1.30 and 1.30-strict tracks are open

addyess:
apparently they are open?  So 👍🏻

PR #1506, opened 2024-04-17
Run coredns charm validation with image from charmhub

addyess:
Summary
------

Remove the hack that installed a multi-arch image from rocks when testing the coredns charm and use the oci image associated with the charm from charmhub

Details
----------------
In order to test on AWS with arm64 cores, I needed to override the `owner` tag used on instances in AWS so they wouldn't be reaped by infra jenkins runs while i'm testing.  I went ahead an parameterized this everywhere


kwmonroe (commented):
Yup, I really like this. The infra reaper has bitten me on more than one occasion because I wasn't paying attention to resources being removed midway through a local test. 

Thanks for this!

PR #1518, opened 2024-06-11
restore ceph testing using ceph-csi charm

addyess:
I've tested this on vsphere and the tests pass as expected.  There's not much difference from before -- just now we're using the ceph-csi charm


kwmonroe (approved):
Lgtm, nice to see bionic/train removed, and +1 on the inclusive update. Thanks!

PR #1534, opened 2024-07-10
remove fruitless assert_hook_occurs_on_all_units

addyess:
`set_config_and_wait` uses `juju-wait` to ensure the config change hooks finish to completion.  There's not a good reason to wait for the `config-change` hook to appear on all units of a certain application.  

Worse yet, this method `assert_hook_occurs_on_all_units` can busy spin forever if it ever misses a config-changed hook. So yeah, lets just toss that

*  **Drive By** : improve logging


asbalderson (approved):
+1

this is great, thank you

PR #1535, opened 2024-07-11
Vault testing no longer requires mysql db, so long as vault is initialized properly

addyess:
## Overview

Vault 1.8/stable channel no longer needs mysql or etcd as backing store for vault, instead it uses raft.  This simplifies the bundle needed to deploy vault and test encryption at rest. 

## Details
* changing the vault-spec to adjust the bundle
* breaks encryption at rest into two tests -- initializing vault, and testing encrypted secrets
* the process for initializing vault is a bit different


jobs/integration/validation.py
```diff
@@ -55,15 +54,9 @@
...
+from juju.model import Model
 from juju.unit import Unit
 
-# Quiet the noise
-logging.getLogger("asyncio").setLevel(logging.INFO)
-logging.getLogger("connector").setLevel(logging.INFO)
-logging.getLogger("websockets.client").setLevel(logging.INFO)
-logging.getLogger("websockets.protocol").setLevel(logging.INFO)
-# bump up juju debug
-logging.getLogger("juju").setLevel(logging.INFO)
```
addyess, line 66: these moved to `conftest.py`

PR #1537, opened 2024-07-24
Adjust keepalived to build with lp-builders

addyess:
Adjust keepalived to be built with launchpad builders

Example build runs)
https://launchpad.net/~addyess/charm-keepalived/+charm/charm-keepalived-akd


louiseschmidtgen (approved):
LGTM

PR #1544, opened 2024-07-25
Swap out juju series for base specification

addyess:
### Overview

Swap out mentions of juju SERIES for juju base

### Details

* in `juju.bash`, create a method called `juju::base::from_series`
* use that method to translate series -> base
* replace calls to bundles to use 
    ```yaml
    series: null
    default-base: $(juju::base::from_series $SERIES)
    ```
* replace `boostrap-series` with `bootstrap-base`

### DIFF
--- a/jobs/arc-conformance/conformance-spec
+++ b/jobs/arc-conformance/conformance-spec
@@ -19,7 +19,7 @@ function juju::bootstrap
     # override because azure needs image-stream=released
     juju bootstrap "$JUJU_CLOUD" "$JUJU_CONTROLLER" \
          --add-model "$JUJU_MODEL" \
-         --bootstrap-series "$SERIES" \
+         --bootstrap-base "$(juju::base::from_series $SERIES)" \
          --force \
          --bootstrap-constraints arch="$ARCH" \
          --model-default test-mode=true \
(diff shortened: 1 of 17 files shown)


kwmonroe (approved):
This LGTM, but i'm unfamiliar with the `default-base` key in overlays.  Is `series: null` still needed when `d-b` is in play?

Regardless, +1 because it's a good transition to get noble support without actually changing our defaults from jammy.

kwmonroe (approved):
This LGTM, but i'm unfamiliar with the `default-base` key in overlays.  Is `series: null` still needed when `d-b` is in play?

Regardless, +1 because it's a good transition to get noble support without actually changing our defaults from jammy.

addyess:
@kwmonroe 

> This LGTM, but i'm unfamiliar with the `default-base` key in overlays. Is `series: null` still needed when `d-b` is in play?
> 
> Regardless, +1 because it's a good transition to get noble support without actually changing our defaults from jammy.

thanks dude. 
[`default-base`](https://juju.is/docs/sdk/bundle.yaml#heading--default-base) is a replacement for `series` in juju bundles
i'm using `series: null` in case a base bundle (kubernetes-core or charmed-kubernetes) tries to specify a series.  Our stable bundles aren't yet updated to remove `series`

PR #1556, opened 2024-08-05
Update build versions of MicroK8s

ktsakalozos:
We update the microk8s builds to include 1.31.

We also install the nvidia gpu drivers because the operator is not able to build and load them on the specific AWS instance we get from juju.


bschimke95 (approved):
LGTM

PR #1559, opened 2024-08-14
Simplify the EKS job so you don't have type in the version information 5 times

addyess:
### Overview

Simplifies running the eks snap builds

### Details
rather than running the job where you set the parameters
* k8s_tag=v1.31.1
* channels=1.31.1/edge,1.31.1/candidate,1.31.1/beta,1.31.1/stable

Now running the job where you set the parameters
* k8s_tag=v1.31.1
* channels=edge,candidate,beta,stable
---

* include all snap risks rather than just edge,beta,candidate
* build the snap channel list using some groovy split, amend, join magic 🪄

### DIFF
--- a/jobs/build-snaps.yaml
+++ b/jobs/build-snaps.yaml
@@ -83,7 +83,7 @@
             Source tag from https://github.com/kubernetes/kubernetes.
       - string:
           name: channels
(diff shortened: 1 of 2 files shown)


mateoflorido (approved):
LGTM. Nice improvement

PR #1560, opened 2024-08-14
Parameterize cloud to run validation-upgrade job

addyess:
### Overview

Jenkins job `validate-charm-release-upgrade` updated to accept a parameter detailing which cloud to tests with

### Details
* jenkins job gets a new string parameter defaulted to vsphere/Boston
* Use of this `cloud` parameter when calling the spec
* updates the spec script to use a 4th parameter -- which cloud to use


kwmonroe (approved):
LGTM, motivated by the upcoming vsphere/Boston move. Thanks for this!!

PR #1562, opened 2024-08-19
Quick skip validation test if models or clouds mismatch

addyess:
### Overview 

* More quickly skip tests if they use the `on_model` test marking
* More quickly skip tests if they use the `clouds` test marking

### Details
Using the pytest feature [`pytest_runtest_setup`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_runtest_setup), we can use the required `model` keyword from config to quickly skip a test if it doesn't match that model name

* This works well with models because the name doesn't have derived from a module level fixture -- but rather test configuration. 
* Removed the "arch" level skipping because it wasn't testing the architecture of the units under test -- but rather the architecture of the machine running the tests. 
* By making the `Tools` object in the test not an async loaded object, it can be constructed and loaded in `pytest_configure`
   * this allows the tests to determine the controller's "cloud" before connecting to the juju model

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -24,7 +24,6 @@ from .utils import (
     asyncify,
     upgrade_charms,
     upgrade_snaps,
-    arch,
     log_snap_versions,
     juju_run,
 )
@@ -343,7 +342,8 @@ async def model(request, tools):
 
 
 @pytest.fixture(scope="module")
-async def k8s_cloud(kubeconfig, tools):
+@pytest.mark.usefixtures("kubeconfig")
+async def k8s_cloud(tools):
     clouds = await tools.run(
         "juju", "clouds", "--format", "yaml", "-c", tools.controller_name
     )
@@ -430,19 +430,6 @@ async def k8s_model(k8s_cloud, tools):
             )
 
 
-@pytest.fixture
-def system_arch():
-    return arch
-
-
-@pytest.fixture(autouse=True)
-def skip_by_arch(request, system_arch):
-    """Skip tests on specified arches"""
-    if request.node.get_closest_marker("skip_arch"):
-        if system_arch in request.node.get_closest_marker("skip_arch").args[0]:
-            pytest.skip("skipped on this arch: {}".format(system_arch))
-
-
 @pytest.fixture(scope="module")
 async def proxy_app(model):
     proxy_app = model.applications.get("squid-forwardproxy")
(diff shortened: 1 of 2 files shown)


jobs/integration/conftest.py
```diff
@@ -52,9 +54,7 @@ def pytest_addoption(parser):
         default=os.environ.get("SERIES", "focal"),
         help="Base series",
     )
-    parser.addoption(
-        "--cloud", action="store", default="aws/us-east-2", help="Juju cloud to use"
-    )
+    parser.addoption("--cloud", action="store", help="Juju cloud to use")
```
addyess, line 57: Stop defaulting the cloud type to 'aws/us-east-2`.  This isn't necessary

jobs/integration/conftest.py
```diff
@@ -143,41 +143,49 @@ def pytest_addoption(parser):
 class Tools:
     """Utility class for accessing juju related tools"""
 
-    def __init__(self, request):
-        self._request = request
+    def __init__(self, config):
+        self._config = config
```
addyess, line 147: construct with the pytest `config` rather than a `request` object

jobs/integration/conftest.py
```diff
@@ -143,41 +143,49 @@ def pytest_addoption(parser):
...
         self.requests = requests
         self.requests_get = asyncify(requests.get)
+        self.run = asyncify(self.exec)
```
addyess, line 150: the existing `tools.run` is an async version of the new `tools.exec` which does away with the async stuff...

jobs/integration/conftest.py
```diff
@@ -143,41 +143,49 @@ def pytest_addoption(parser):
...
+    @cached_property
+    def cloud(self):
+        if _it := self._config.getoption("--cloud"):
+            return _it
+        controller_data, _ = self.exec(
+            "juju", "show-controller", self.controller_name, "--format", "yaml"
         )
+        controller_infos = yaml.safe_load(controller_data)
+        controller_info, *_ = controller_infos.values()
+        return controller_info["details"]["cloud"]
```
addyess, line 188: the cloud field can be loaded from the existing **required** controller name argument.

jobs/integration/conftest.py
```diff
@@ -191,9 +199,9 @@ def juju_base(self, series):
         }
         return f"--base={mapping[series]}"
 
-    async def run(self, cmd: str, *args: str, stdin=None, _tee=False):
+    def exec(self, cmd: str, *args: str, stdin=None, _tee=False):
```
addyess, line 202: `async subprocess` is actually really hard -- this method actually gets much simpler

jobs/integration/conftest.py
```diff
@@ -304,9 +289,7 @@ async def fast_forward(
 
 @pytest.fixture(scope="module")
 async def tools(request):
-    tools = Tools(request)
-    await tools._load()
-    return tools
+    yield request.config.test_tools
```
addyess, line 292: the old fixture now just has to yield the existing tools from the `request.config`.  See below in `pytest_configure`

jobs/integration/conftest.py
```diff
@@ -719,6 +692,17 @@ def pytest_metadata(metadata):
         )
 
 
+def pytest_runtest_setup(item):
+    """Called to perform the setup phase for a test item."""
+    skip_by_model(item)  # skip tests if model marking on test mismatches
+    skip_by_cloud(item)  # skip tests if cloud marking on test mismatches
```
addyess, line 698: handles quick skipping of the tagged test by model or cloud

jobs/integration/conftest.py
```diff
@@ -719,6 +692,17 @@ def pytest_metadata(metadata):
...
+
+
+def pytest_configure(config):
+    config.test_tools = Tools(config)
+    config.test_tools._load()
```
addyess, line 703: Load the tools for the validation tests into the config's global space

kwmonroe (approved):
Yup, this LGTM and is a welcomed fix to better include/skip bits on non-pytest-running arches and for non-gpu capable models (for stuff like nvidia tests).

PR #1563, opened 2024-08-20
Readd testing nginx-ingress with microbot test

addyess:
### Overview

Test the kubernetes-worker's nginx-ingress using the microbot appliance.

### Details
* Adds a template for the microbot deployment
* Adds a test to setup and teardown the deployment and reach through the ingress

### DIFF
--- a/jobs/integration/templates/microbot.yaml
+++ b/jobs/integration/templates/microbot.yaml
@@ -1,0 +1,66 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
+  labels:
+    app: microbot
+  name: microbot
+spec:
+  replicas: {{ replicas }}
+  selector:
+    matchLabels:
+      app: microbot
+  strategy: {}
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        app: microbot
+    spec:
+      containers:
+      - image: {{ registry|default("docker.io") }}/cdkbot/microbot-{{ arch }}:latest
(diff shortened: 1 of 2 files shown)


jobs/integration/templates/microbot.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
```
addyess, line 1: lifted this file straight from the reactive archive of the kubernetes-worker charm
https://github.com/charmed-kubernetes/charm-kubernetes-worker/blob/reactive-archive/templates/microbot-example.yaml

jobs/integration/validation.py
```diff
@@ -2617,7 +2645,8 @@ async def test_cinder(model, tools):
 
 
 @pytest.mark.clouds(["openstack"])
-async def test_octavia(model, tools, teardown_microbot):
+@pytest.mark.usefixtures("teardown_microbot")
+async def test_octavia(model, tools):
```
mateoflorido, line 2649:
```suggestion
@pytest.mark.usefixtures("teardown_microbot")
async def test_octavia(model, tools):
# TODO: Update the microbot deployment mechanism since the action has been deprecated.
```

jobs/integration/templates/microbot.yaml
addyess: rather than adding it back to the kubernetes-worker charm -- just use this sample to start the microbot in tests

mateoflorido (approved):
LGTM

PR #1580, opened 2024-10-09
Update snapcraft go version based on changes to upstream's `.go-version`

addyess:
### Overview
Build snaps according to a rolling upstream go-version

### Details
Upstream https://github.com/kubernetes/kubernetes began to embed the required go-version for a specific branch or tag using the `.go-version` file.  It's recommended to use this version of go to build the snaps. 

Charmed Kubernetes has been keeping an internal map which can get stale -- rather than change it -- we can continue to use it if the supported branch doesn't contain the .go-version file.  

* the sync-snaps job will now confirm the correct go-version in the snapcraft yaml is used.

PR #1587, opened 2024-10-23
Track repo differences with git status rather than git diff

addyess:
### Details
When new patch releases of upstream kubernetes are published as tags, we store new branches internally on launchpad, each with it's own `snapcraft.yaml` to build the various snaps. 

https://github.com/charmed-kubernetes/jenkins/pull/1580 introduced a change to only push these new branches when a difference was detected.  However using `git diff` on a branch where the only change is a new file results in an empty set of CHANGED files.  

Switching to git status shows the new files:

example)
```sh
⚡ git diff
⚡ git status
On branch v1.29.10
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	snapcraft.yaml

nothing added to commit but untracked files present (use "git add" to track)
⚡ git status --porcelain                                                                       
?? snapcraft.yaml
```

### Details
* adjust the check from git diff to git status which returns a list of changed files and how they were changed

### DIFF
--- a/cilib/git.py
+++ b/cilib/git.py
@@ -5,7 +5,7 @@ import logging
 import sh
 import re
 from subprocess import run
-from typing import Dict
+from typing import Dict, List
 from .github_api import Repository
 import requests
 import requests.auth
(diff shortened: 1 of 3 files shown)


mateoflorido (approved):
LGTM!

mateoflorido (approved):
Nice. Thank you for taking care of the moved files in case we need them in the future

PR #1607, opened 2025-01-08
Revert "Update enums for 1.32 GA"

bschimke95:
Reverts charmed-kubernetes/jenkins#1605

The stable branches should not be populated before solQA approval.


addyess:
@bschimke95 was right, i was wrong.  We didn't need this

PR #1612, opened 2025-02-26
chore: Update k8s version to 1.32/stable

HomayoonAlimohammadi:
### Overview
This PR bumps K8s version to 1.32, adds 1.33 and removes older versions where necessary according to the ["Confirm snap promotion"](https://github.com/charmed-kubernetes/jenkins/blob/main/docs/releases/stable/index.md#confirm-snap-promotion-from-1xxrisk-to-latestrisk) step.


addyess (approved):
Right.  I like this... there may be more bumps from 1.31 -> 1.32 but this is a good start

PR #1614, opened 2025-03-03
Stop testing Charmed Kubernetes against focal

addyess:
### Overview
No longer test charmed kubernetes against focal

### Details
* Ends testing on python38
* migrates default series from focal to jammy
* drops nightly testing of `focal` without yet adding `noble` testing

---

- [x] Needs `jjb` after merge

### DIFF
--- a/jobs/validate.yaml
+++ b/jobs/validate.yaml
@@ -67,9 +67,6 @@
       - jammy:
           hour: '0' # Midnight
           dow: '1,3,5'  # MWF
-      - focal:
-          hour: '12' # Noon
-          dow: '2,4,6'  # TuThSa
     charm-channel:
       - edge:
           snap_versions:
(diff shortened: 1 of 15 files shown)


jobs/validate.yaml
```diff
@@ -67,9 +67,6 @@
       - jammy:
           hour: '0' # Midnight
           dow: '1,3,5'  # MWF
```
mateoflorido, line 69: Should we add jobs for `noble`? I know these should be added once we have charms that support `noble`, right?

addyess: Not yet, not all the k8s charms will yet support noble.  Only some of them.  All we can test with is jammy until we migrate all of them.

mateoflorido (approved):

PR #1621, opened 2025-03-20
Make aws heavy tests run on demand rather than scheduled

addyess:
The following tests were somewhat costly and are perpetually in a failed state
* validate-ck-aws-iam
* validate-localhost
* validate-nvidia

Jira cards have been created to address 2 of these tests going forward after which they can be re-enabled as schedule tests runs. 

For now, we can just disable the scheduled test runs


mateoflorido (approved):
LGTM!

PR #1623, opened 2025-03-31
Bugfix-release-upgrade should have a configurable snap offices channels

addyess:
When running a bugfix-release-upgrade job, we can't just test two specific snap versions, we must calculate which snaps to deploy based on the candidate charm channel. 

We're trying to release 1.xx/candidate charm works with the following snap releases
* 1.(xx-0)/stable
* 1.(xx-1)/stable

The test must upgrade 
charm: `1.xx/stable` -> `1.xx/candidate`
snap: (`$snap_version`) -> `1.xx/candidate`

where the `snap_version` is calculated using the prior tracks

---

- [ ] Needs `jjb` after merge

### DIFF
--- a/jobs/release.yaml
+++ b/jobs/release.yaml
@@ -80,10 +80,16 @@
             - runner-validate
       - axis:
           type: user-defined
-          name: snap_version
+          name: prior_snap_offset
+          description: |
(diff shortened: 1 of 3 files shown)


mateoflorido (approved):
LGTM!

PR #1624, opened 2025-04-01
Handle snap upgrade requirements across multiple ck releases

addyess:
### Overview
Resolves a few issues with Bugfix Upgrade testings across multiple releases (1.28 -> 1.32)

Starting in 1.29, changing snaps from `1.xx/stable` -> `1.xx/candidate` didn't actually block the charm, so logic was added to the tests to skip running the upgrade action. This is still a requirement for 1.28 however.  

So, lets just have the test run the action if the charm says it needs the action run. 

### Details
* fix logic bug on when to upgrade charms and snaps
* Reinstate focal testing for the time being on bugfix releases
* drop juju 3.1 for nightly tests runs

---

- [x] Needs `jjb` after merge

### DIFF
--- a/jobs/integration/conftest.py
+++ b/jobs/integration/conftest.py
@@ -333,7 +333,7 @@ async def model(request, tools):
         await tools.juju_wait()
         upgrade_snap_channel = request.config.getoption("--upgrade-snap-channel")
         upgrade_charm_channel = request.config.getoption("--upgrade-charm-channel")
-        if not upgrade_snap_channel and upgrade_charm_channel:
+        if not (upgrade_snap_channel and upgrade_charm_channel):
             raise Exception(
                 "Must have both snap and charm upgrade "
                 "channels set to perform upgrade prior to validation test."
--- a/jobs/integration/utils.py
+++ b/jobs/integration/utils.py
@@ -118,55 +118,41 @@ async def upgrade_charms(model, channel, tools):
     await tools.juju_wait()
 
 
-async def upgrade_snaps(model, channel, tools):
-    for app_name, blocking in {
-        "kubernetes-control-plane": True,
-        "kubernetes-worker": True,
-        "kubernetes-e2e": False,
-    }.items():
+async def upgrade_snaps(model: Model, channel, tools):
+    for app_name in [
+        "kubernetes-control-plane",
+        "kubernetes-worker",
+        "kubernetes-e2e",
+    ]:
         app = model.applications.get(app_name)
         # missing applications are simply not upgraded
         if not app:
             continue
 
-        config = await app.get_config()
         # If there is no change in the snaps skipping the upgrade
+        config = await app.get_config()
         current_channel = config["channel"]["value"]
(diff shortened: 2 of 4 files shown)


jobs/integration/conftest.py
```diff
@@ -333,7 +333,7 @@ async def model(request, tools):
         await tools.juju_wait()
         upgrade_snap_channel = request.config.getoption("--upgrade-snap-channel")
         upgrade_charm_channel = request.config.getoption("--upgrade-charm-channel")
-        if not upgrade_snap_channel and upgrade_charm_channel:
+        if not (upgrade_snap_channel and upgrade_charm_channel):
```
addyess, line 336: the boolean logic here was just, incorrect.

jobs/integration/utils.py
```diff
@@ -118,55 +118,41 @@ async def upgrade_charms(model, channel, tools):
...
-        "kubernetes-control-plane": True,
-        "kubernetes-worker": True,
-        "kubernetes-e2e": False,
-    }.items():
+async def upgrade_snaps(model: Model, channel, tools):
+    for app_name in [
+        "kubernetes-control-plane",
+        "kubernetes-worker",
+        "kubernetes-e2e",
+    ]:
```
addyess, line 126: we don't need to know whether or not to expect the applications to block when they undergo a snap upgrade... all we really need to check is 
1) Do the units go idle after the config change (not necessarily active/idle)
2)  is the unit blocked specifically because it needs a manual upgrade

if so -- run the upgrade action.

jobs/integration/utils.py
```diff
@@ -118,55 +118,41 @@ async def upgrade_charms(model, channel, tools):
...
-                    ):
-                        break
-                    await asyncio.sleep(3)
-                else:
-                    raise asyncio.TimeoutError(
-                        "Unable to find blocked status on unit {0} - {1} {2}".format(
-                            unit.name, unit.workload_status, unit.agent_status
-                        )
-                    )
+        await model.wait_for_idle(apps=[app_name])
```
addyess, line 140: The above look was overly complicated.  It wanted to assume that these applications would block in certain situations, and not block in others.  

Let's replace this with some simple checks
1) are the units in this application idle (not active/idle)
2) do any "Need a manual upgrade", if so run the upgrade action

jobs/release.yaml
```diff
@@ -90,10 +90,12 @@
           values:
             - 0
```
addyess, line 91: this performs a snap upgrade from 1.xx/stable to 1.xx/candidate
* can the current stable snaps be upgraded to the potential NEXT stable snaps

jobs/release.yaml
```diff
@@ -90,10 +90,12 @@
...
             - 1
+            - 2
```
addyess, line 93: this performs a snap upgrade from 
* 1.(xx-1)/stable to 1.xx/candidate
* 1.(xx-2)/stable to 1.xx/candidate

PR #1646, opened 2025-08-05
Pause building a set of charms that have fallen out of test matrix

addyess:
The following charms did not get a 1.33/stable release and we should likely pause their work since their test-coverage isn't functional.  

* multus
* sriov-cni
* sriov-network-device-plugin
* gatekeeper-controller-manager
* gatekeeper-audit


HomayoonAlimohammadi (approved):
Let's take these off our plates for now. Thanks @addyess! LGTM

addyess:
The decision was made to carry over the built charms from the 1.32/stable release channel via promotion.

PR #1648, opened 2025-08-05
fix: use `SERIES` based on the Control Plane machines

mateoflorido:
## Overview
Determine the Ubuntu Series based on the Control Plane machine units.


addyess (commented):
Be less defensive... (or more assertive)

addyess (approved):
Nice, can't wait to see the magic 🪄

PR #1649, opened 2025-08-05
Load machine series from a given machine's base

addyess:
### Overview

A juju machine's "series" may be missing, in which case we should load the series from the application's base.

### DIFF
--- a/cilib/enums.py
+++ b/cilib/enums.py
@@ -2,6 +2,7 @@
 
 from enum import Enum
 from functools import total_ordering
+from typing import Tuple
 from pathlib import Path
 import yaml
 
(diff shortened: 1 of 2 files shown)


cilib/enums.py
```diff
@@ -41,6 +41,25 @@ class Series(Enum):
...
+            raise ValueError(
+                f"Invalid base format: '{base}' (expected 'distro@version')"
+            )
+
+        if distro != "ubuntu":
+            raise ValueError(f"Unsupported base '{base}' for Series lookup")
+
+        for member in cls:
+            if version.endswith(member.value):
+                return member
```
mateoflorido, line 59: Is it possible to do something like this?
```suggestion
        version_tuple = tuple(map(int, version.split(".")))
        for member in cls:
            if member._version_tuple() == version_tuple:
                return member
```

mateoflorido (commented):
Quick question below 👇🏻

mateoflorido (approved):
LGTM! Thanks @addyess

PR #1680, opened 2026-04-30
fix: remove `juju-pipeline` dependency

mateoflorido:
## Overview

Remove dependency on https://github.com/adam-stokes/juju-jenkins-pipeline

---

- [ ] Needs `jjb` after merge


ethandcosta (approved):
LGTM

